PR #1, opened 2022-11-04
Optimize accounting logic and storage slots in Kitty.sol

haydenshively:


core/src/InterestModel.sol
```diff
@@ -4,15 +4,16 @@ pragma solidity ^0.8.15;
...
 contract InterestModel {
-    function getAccrualFactor(uint256 elapsedTime, uint256 utilization) external returns (uint256 accrualFactor) {
+    // TODO will need both view and non-view versions of this
+    function getAccrualFactor(uint256 elapsedTime, uint256 utilization) external view returns (uint256 accrualFactor) {
         // TODO use utilization to update PID controller
         // TODO each Kitty gets needs its own InterestModel in order for PID controller state to work well. Could still point to single proxy for logic though.
 
         // If utilization > 50%, use 4% APY. 2% APY otherwise.
-        uint256 interestRate = utilization > 0.5e18 ? 1.24e9 : 6.27e8; // ((1 + r) ^ (1 / SECONDS_IN_YEAR) - 1) * 1e18
+        uint256 interestRate = utilization > 0.5e18 ? 1.24e3 : 6.27e2; // ((1 + r) ^ (1 / SECONDS_IN_YEAR) - 1) * 1e12
```
IanWoodard, line 13: I am still a bit confused about these numbers.

core/src/Kitty.sol
```diff
@@ -5,183 +5,259 @@ import {ERC20, SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
...
-            _inventory,
-            _inventory - accruedInterest / 8 // `8` indicates a 12.5% reserve factor
+            slot0.totalSupply,
+            inventory,
+            inventory - (borrowsNew - borrowsOld) / 8 // `8` indicates a 1/8=12.5% reserve factor
         );
-        _mint(treasury, newTotalSupply - totalSupply);
+        if (newTotalSupply != slot0.totalSupply) {
+            slot0.totalSupply = uint112(newTotalSupply); // TODO safe casting
+            _unsafeMint(TREASURY, newTotalSupply - slot0.totalSupply);
```
IanWoodard, line 199: Wouldn't the second argument always be zero?

core/src/Kitty.sol
```diff
@@ -5,183 +5,259 @@ import {ERC20, SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
...
+        }
+
+        return (slot0, slot1, inventory);
+    }
 
-        return (_packedSlot, newTotalSupply, _inventory);
+    function _save(Slot0 memory slot0, Slot1 memory slot1) private {
+        totalSupply = slot0.totalSupply;
+        lastBalance = slot0.lastBalance;
+        lastAccrualTime = slot0.lastAccrualTime; // could just set to block.timestamp
```
IanWoodard: is there any benefit in doing it one way versus the other?

core/src/Kitty.sol
```diff
@@ -5,183 +5,259 @@ import {ERC20, SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
...
         uint256 newTotalSupply = FullMath.mulDiv(
-            totalSupply,
-            _inventory,
-            _inventory - accruedInterest / 8 // `8` indicates a 12.5% reserve factor
+            slot0.totalSupply,
+            inventory,
+            inventory - (borrowsNew - borrowsOld) / 8 // `8` indicates a 1/8=12.5% reserve factor
         );
-        _mint(treasury, newTotalSupply - totalSupply);
+        if (newTotalSupply != slot0.totalSupply) {
```
IanWoodard, line 198: When would the above not be true?

core/src/KERC20.sol
```diff
@@ -0,0 +1,198 @@
...
+
+    /*//////////////////////////////////////////////////////////////
+                        INTERNAL MINT/BURN LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev You must do `totalSupply += amount` separately. Do so in a checked context.
+    function _unsafeMint(address to, uint256 amount) internal virtual {
+        unchecked {
+            balanceOf[to] += amount;
+        }
```
IanWoodard, line 187: Why do we want to wrap on overflow rather than revert?

IanWoodard: Also, why does the comment above say to do it in a checked context?

core/src/KERC20.sol
```diff
@@ -0,0 +1,198 @@
...
+    function approve(address spender, uint256 amount) public virtual returns (bool) {
+        allowance[msg.sender][spender] = amount;
+
+        emit Approval(msg.sender, spender, amount);
+
+        return true;
+    }
+
+    function transfer(address to, uint256 amount) public virtual returns (bool) {
+        balanceOf[msg.sender] -= amount;
```
IanWoodard, line 77: This would revert in the case of underflow, correct?

core/napkin.py
```diff
@@ -0,0 +1,143 @@
...
+    inventory = one_unit
+    total_supply = inventory * Decimal(10 ** total_supply_precision)
+    interest = Decimal(1)
+    while True:
+        interest += 1
+        total_supply_new = total_supply * (inventory + interest) / (inventory + interest - math.floor(interest / RESERVE_DIVISOR))
+        if math.floor(total_supply_new) > total_supply:
+            print((total_supply, inventory, interest))
+            break
+    print(interest)
```
IanWoodard: why are we printing instead of asserting?

core/src/Kitty.sol
```diff
@@ -5,183 +5,259 @@ import {ERC20, SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
...
+        slot0.lastBalance -= uint112(amount); // TODO safe casting
+        ASSET.safeTransfer(to, amount);
+        
+        _save(slot0, slot1);
     }
 
-    function repay(uint128 amount) external onlyMarginAccount {
-        // Poke (includes reentrancy guard)
-        (PackedSlot memory _packedSlot, , ) = accrueInterest();
+    function repay(uint256 amount, address to) external {
```
IanWoodard, line 145: Would it be smart to check to see if the to address is what we are expecting?

IanWoodard (commented):
Left some initially thoughts

core/src/Kitty.sol
haydenshively: Oops yes good catch, ordering of lines 200 and 201 should be flipped.

core/src/Kitty.sol
haydenshively: Yes, better to set to block.timestamp. Planning to do this soon; should've put TODO by this comment.

core/src/InterestModel.sol
haydenshively: According to the comments at the end of the lines, 4% APY would be:
= ((1+0.04)^0.000000031709792 - 1) * 1e12
= (1.000000001243681 - 1) * 1e12
= 1243.68
= 1.244e3

Does that make sense?

core/src/Kitty.sol
haydenshively: Context:
`borrowsOld = borrowBase * borrowIndexOld / SCALER`
`borrowsNew = borrowBase * borrowIndexNew / SCALER`

Though `borrowIndexNew` is always greater than `borrowIndexOld`, we're dividing by SCALER to get `borrowsOld` and `borrowsNew`. This decrease in precision implies that `borrowsNew > borrowsOld` may not always be true.

In order for `newTotalSupply != slot0.totalSupply`, not only do we need `borrowsNew > borrowsOld`, but we _also_ need `floor((borrowsNew - borrowsOld) / 8)` to be nonzero. So that delta needs to be at least 8 to change total supply.

In practice, if the Kitty is managing more than ~$100, this will always be true. But on the off chance that a token has weird decimals, or our interest rates are really low, or nobody has borrowed anything yet... checking whether total supply has changed can save us gas.

core/src/KERC20.sol
haydenshively: Correct

core/src/KERC20.sol
haydenshively: This file was taken from https://github.com/transmissions11/solmate/blob/main/src/tokens/ERC20.sol and lightly modified. The original `_mint` function looked like this:
```Solidity
function _mint(address to, uint256 amount) internal virtual {
    totalSupply += amount;

    // Cannot overflow because the sum of all user
    // balances can't exceed the max uint256 value.
    unchecked {
        balanceOf[to] += amount;
    }

    emit Transfer(address(0), to, amount);
}
```

In this version, `_mint` both reads and writes from whatever storage slot `totalSupply` is in. This is gassy! In my version, `totalSupply` is stored as a `uint112` instead of `uint256`, allowing me to pack other variables into the same storage slot:
- `uint112 totalSupply`
- `uint112 lastBalance`
- `uint32 lastAccrualTime`
Note that these sum to 256 bits, which means I'm using every last bit of the slot.

So, getting back to `_mint`: If we keep the `totalSupply += amount` inside this function, it would copy the entire slot from storage, update the `totalSupply` portion in memory, and then copy the updated one back to storage. It'd be far better to only write to the packed slot once I have **all** the updated info -- `totalSupply, `lastBalance` and `lastAccrualTime`. To make that happen, I remove the `totalSupply` increment from here and do it elsewhere. BUT in order to avoid overflows in the `unchecked` balanceOf block, it's important that regardless of where I increment totalSupply, I do it in a checked context.

core/src/Kitty.sol
haydenshively: Well for borrowing, we definitely need to make sure that only official MarginAccount instances have access. But for repaying, I think we can open it up to anyone. If you want to repay Bob's debt for him, go ahead! The protocol doesn't care. The more practical use-case is if someone's MarginAccount is about to get liquidated. Rather than going through that to pay off debt, they can call this function from **any address** that happens to have the requisite funds. This may be more gas efficient for them.

PR #2, opened 2022-11-16
Configure Lender.sol to be clonable with immutable args

haydenshively:
This chart shows the gas costs associated with the current testnet contracts:
image

Since we're now routing all `Lender` calls through proxies, one may expect gas costs to increase. But because of our other optimizations, they've actually decreased:
image

One note: In the live version, ERC20 costs are automatically included in `deposit` because it uses the "pull" architecture. Now we're using push, so we have to add those costs on top of the gas report. As such, the median gas usage of `deposit` should be around 16,000 not 10,000.

_looking at median values_
`accrueInterest`: 80% better (looking at min for this; median doesn't mean much since there were only 2 calls)
`deposit`: 77% better
`withdraw`: 64% better
`borrow`: 53% better
`repay`: about the same


core/test/Kitty.t.sol
```diff
@@ -3,11 +3,10 @@ pragma solidity ^0.8.15;
...
 import "forge-std/Test.sol";
 
-import {ERC20} from "solmate/tokens/ERC20.sol";
-
-import "src/InterestModel.sol";
 import "src/Lender.sol";
 
+import {deploySingleLender} from "./Utils.sol";
+
 contract LenderTest is Test {
```
IanWoodard, line 10: Should this file be renamed to Lender.t.sol?

core/src/Factory.sol
```diff
@@ -40,8 +46,18 @@ contract Factory {
 
         // TODO this implies that lending pairs are fee-tier specific. does it make sense to combine fee tiers?
         //      if so, margin account Uniswap liquidity readers will have to change.
-        Lender lender0 = new Lender{salt: keccak256(abi.encode(_pool))}(asset0, INTEREST_MODEL, address(this));
-        Lender lender1 = new Lender{salt: keccak256(abi.encode(_pool))}(asset1, INTEREST_MODEL, address(this));
+        bytes32 salt = keccak256(abi.encode(_pool));
```
IanWoodard, line 49: Is this so we avoid hashing twice?

core/src/Factory.sol
```diff
@@ -40,8 +46,18 @@ contract Factory {
...
+        Lender lender1 = Lender(lenderImplementation.cloneDeterministic({
+            salt: salt,
+            data: abi.encodePacked(address(asset1))
+        }));
+
+        lender0.initialize();
+        lender1.initialize();
 
         getMarket[_pool] = Market(lender0, lender1);
         emit CreateMarket(_pool, lender0, lender1);
```
IanWoodard, line 63: I know you didn't change the stuff down here but should line 71 say
"a WETH/USDC margin account and a WETH/WBTC margin account shouldn't be able to borrow from the same WETH lender" instead of what it currently says.

IanWoodard (approved):
LGTM

core/test/Kitty.t.sol
haydenshively: Yes. I'll take care of this in the next PR, since there's still some more renaming to do

core/src/Factory.sol
haydenshively: Yep!

PR #4, opened 2022-12-09
Adding BorrowManager

IanWoodard:
Adding BorrowManager to enable non-leveraged borrowing for the front end.

### DIFF
--- a/periphery/src/BorrowManager.sol
+++ b/periphery/src/BorrowManager.sol
@@ -59,11 +59,9 @@ contract BorrowManager is IManager {
                 }
             } else if (action == 2) {
                 if (amount0 != 0) {
-                    ERC20 token0 = account.TOKEN0();
                     token0.safeTransferFrom(address(account), owner, amount0);
                 }
                 if (amount1 != 0) {
-                    ERC20 token1 = account.TOKEN1();
                     token1.safeTransferFrom(address(account), owner, amount1);
                 }
             }
--- a/periphery/src/BorrowManager.sol
+++ b/periphery/src/BorrowManager.sol
@@ -36,8 +36,8 @@ contract BorrowManager is IManager {
             (uint8[], uint256[], uint256[])
         );
         address owner = account.owner();
-        ERC20 token0 = account.TOKEN0();
-        ERC20 token1 = account.TOKEN1();
+        ERC20 token0;
+        ERC20 token1;
 
         require(actions.length == amounts0.length && actions.length == amounts1.length, "Aloe: bad data");
         for (uint256 i; i < actions.length; i++) {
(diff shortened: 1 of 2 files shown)


periphery/src/BorrowManager.sol
```diff
@@ -0,0 +1,67 @@
...
+contract BorrowManager is IManager {
+    using SafeTransferLib for ERC20;
+
+    Factory public immutable FACTORY;
+
+    constructor(Factory _factory) {
+        FACTORY = _factory;
+    }
+
+    /* solhint-disable code-complexity */
```
haydenshively, line 20: Can you move this disable to a more specific location, and re-enable the code-complexity check after the problematic section of code?

periphery/src/BorrowManager.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * @notice This should be called by a borrower during its modify function
+     * @param data actions, amount0, amount1
```
haydenshively: I'd recommend being more specific here. Something like the following (depends on my other comments):

```
@param data Obtained from `abi.encode(actions, amount0, amount1)`. User should pass this in when calling `borrower.modify`
```

periphery/src/BorrowManager.sol
```diff
@@ -0,0 +1,67 @@
...
+     * @dev actions: 0 = borrow, 1 = repay, 2 = withdraw
+     * @return positions
+     * @return includeLenderReceipts
+     */
+    function callback(bytes calldata data)
+        public
+        returns (Uniswap.Position[] memory positions, bool includeLenderReceipts)
+    {
+        require(FACTORY.isBorrower(msg.sender), "Aloe: bad account");
+        Borrower account = Borrower(msg.sender);
```
haydenshively, line 34: style nit: I'd encourage the use of more newlines to make this easier to read

periphery/src/BorrowManager.sol
```diff
@@ -0,0 +1,67 @@
...
+        (uint8[] memory actions, bytes[] memory args) = abi.decode(data, (uint8[], bytes[]));
+        require(actions.length == args.length, "Aloe: bad data");
+        for (uint256 i; i < actions.length; i++) {
+            uint8 action = actions[i];
+            (uint256 amount0, uint256 amount1) = abi.decode(args[i], (uint256, uint256));
```
haydenshively: Since the structure of `args` is the same for all indices, you could avoid nested encoding. For example:
```
(uint8[] memory actions, uint256[] amounts0, uint256[] amounts1) = abi.decode(data, (uint8[], uint256[], uint256[]));
```

Maybe try both and see which one is more gas efficient?

periphery/src/BorrowManager.sol
```diff
@@ -0,0 +1,67 @@
...
+            if (action == 0) {
+                account.borrow(amount0, amount1, account.owner());
+            } else if (action == 1) {
+                if (amount0 != 0) {
+                    ERC20 token0 = account.TOKEN0();
```
haydenshively: We're spending extra gas every time we fetch `TOKEN0` from `account`. Maybe move these outside of the for loop, and only fetch them if the local variable hasn't yet been set, i.e.:

`if (token0 == address(0)) token0 = account.TOKEN0();`

periphery/src/BorrowManager.sol
```diff
@@ -0,0 +1,67 @@
...
+                    Lender lender0 = account.LENDER0();
+                    token0.safeTransferFrom(account.owner(), address(lender0), amount0);
```
haydenshively: I would just store `account.owner()` in a local address variable outside of the for loop. (gas savings)

periphery/src/BorrowManager.sol
IanWoodard: Wouldn't the amounts need to be memory?

periphery/src/BorrowManager.sol
IanWoodard: Why do we need the additional check?

periphery/src/BorrowManager.sol
IanWoodard: Will do, just a heads up, I used the disable from FrontendManager

periphery/src/BorrowManager.sol
IanWoodard: I see, added.

haydenshively (approved):
Looks great!

PR #5, opened 2022-12-13
Add more Lender tests and use programmatic forks

haydenshively:


core/test/libraries/Volatility.t.sol
```diff
@@ -45,7 +45,7 @@ contract VolatilityTest is Test {
             Volatility.FeeGrowthGlobals(
                 1501968291161650295867029090958139,
                 527315901327546020416261134123578344760082,
-                uint32(block.timestamp)
+                uint32(1660599905)
```
IanWoodard, line 48: why did this change?

core/test/Lender.t.sol
```diff
@@ -3,21 +3,170 @@ pragma solidity ^0.8.15;
...
+        deal(address(asset), address(lender), amount);
+
+        assertEq(lender.balanceOfUnderlying(address(12345)), balance);
+    }
+
+    function test_cannotDepositWithoutERC20Transfer(uint112 amount, address to) public {
+        if (amount == 0) amount++;
+
+        deal(address(asset), address(lender), amount - 1);
+        vm.expectRevert(bytes(""));
```
IanWoodard, line 124: We should specify the revert

core/test/Lender.t.sol
```diff
@@ -3,21 +3,170 @@ pragma solidity ^0.8.15;
...
+        lender.whitelist(borrower);
+        assertEq(lender.borrows(borrower), 1);
+
+        // Whitelisting does not increase debt
+        assertEq(lender.borrowBalanceStored(borrower), 0);
+
+        // Whitelisting cannot erase debt
+        if (value == 0) value++;
+        stdstore.target(address(lender)).sig("borrows(address)").with_key(borrower).depth(0).checked_write(value);
+        vm.expectRevert(bytes(""));
```
IanWoodard, line 45: we should specify the revert

IanWoodard (approved):
LGTM

core/test/libraries/Volatility.t.sol
haydenshively: Honestly I'm not entirely sure, but I'm pretty sure foundry doesn't properly update block.timestamp in the programmatic forks.

core/test/Lender.t.sol
haydenshively: This is as specific as we can get right now, since no real error message is emitted. `bytes("")` at least tells it to expect an empty reason string, as opposed to _any_ reason string.

core/test/Lender.t.sol
haydenshively: See above

PR #6, opened 2022-12-14
Add gas efficient referral program (couriers get cut of interest)

haydenshively:


core/.gas-snapshot
```diff
@@ -1,12 +1,12 @@
...
-FactoryGasTest:test_createMarket() (gas: 3919974)
-LenderGasTest:test_accrueInterest() (gas: 36818)
+FactoryGasTest:test_createBorrower() (gas: 154894)
+FactoryGasTest:test_createMarket() (gas: 3919934)
+LenderGasTest:test_accrueInterest() (gas: 36937)
 LenderGasTest:test_borrow() (gas: 41864)
-LenderGasTest:test_deposit() (gas: 43720)
-LenderGasTest:test_redeem() (gas: 43345)
-LenderGasTest:test_repay() (gas: 44956)
+LenderGasTest:test_deposit() (gas: 43883)
```
haydenshively, line 10: (Adding features increases gas)

core/.storage-layout
```diff
@@ -24,9 +24,11 @@ forge inspect --pretty src/Lender.sol:Lender storage-layout
 |------------------------+-------------------------------------------------+------+--------+-------+-----------------------|
 | nonces                 | mapping(address => uint256)                     | 7    | 0      | 32    | src/Lender.sol:Lender |
 |------------------------+-------------------------------------------------+------+--------+-------+-----------------------|
-| interestModel          | contract InterestModel                          | 8    | 0      | 20    | src/Lender.sol:Lender |
+| couriers               | mapping(uint32 => struct Ledger.Courier)        | 8    | 0      | 32    | src/Lender.sol:Lender |
```
haydenshively, line 27: Inserted couriers mapping at slot 8, so other storage gets shifted down

core/src/Ledger.sol
```diff
@@ -112,15 +127,20 @@ contract Ledger {
         }
     }
 
+    /// @notice The number of shares held by `account`
+    function balanceOf(address account) external view returns (uint256) {
+        return balances[account] % Q112;
```
haydenshively, line 132: mod basically just chops off higher order bits

core/src/Lender.sol
```diff
@@ -98,16 +106,16 @@ contract Lender is Ledger {
...
-        _unsafeBurn(owner, shares);
+        // Burn shares and (if applicable) handle courier accounting
+        amount = _unsafeBurn(owner, shares, amount);
         unchecked {
             cache.totalSupply -= shares;
         }
 
+        // Transfer tokens
+        cache.lastBalance -= amount;
+        asset().safeTransfer(recipient, amount);
```
haydenshively, line 117: Moved this logic down below the `_unsafeBurn` since `_unsafeBurn` now updates `amount`

core/src/Ledger.sol
IanWoodard: Why would we want to do this?

IanWoodard: Oh, i see

core/src/Lender.sol
```diff
@@ -330,22 +322,89 @@ contract Lender is Ledger {
...
+        unchecked {
+            // From most to least significant...
+            // -------------------------------
+            // | courier id       | 32 bits  |
+            // | user's principle | 112 bits |
+            // | user's balance   | 112 bits |
+            // -------------------------------
+            uint256 data = balances[from];
+            require(data >> 224 == 0);
+            require(shares <= data % Q112);
```
IanWoodard, line 338: Couldn't you combine these?

core/src/Lender.sol
```diff
@@ -330,22 +322,89 @@ contract Lender is Ledger {
...
+
+            if (data >> 224 != 0) {
+                // Keep track of principle iff courier deserves credit
+                require(amount + ((data >> 112) % Q112) < Q112);
+                data += amount << 112;
+            }
+
+            // Keep track of balance regardless of courier.
+            // Since `totalSupply` fits in uint112, the user's balance will too. No need to check here.
+            balances[to] = data + shares;
```
IanWoodard, line 366: unchecked?

core/src/Ledger.sol
IanWoodard: So this is just the last 112 bits?

haydenshively: Correct. The maximum `totalSupply` of banknotes (receipt tokens) is `2**112 - 1`, so each individual user's balance is guaranteed to fit in 112 bits.

core/src/Lender.sol
haydenshively: Yes, but once I add custom error codes, each failure will report a different error. So it makes sense to keep them separate imo.

core/src/Lender.sol
haydenshively: Yes. Though this function only touches the mapping, anytime we `_unsafeMint`, we also need to increment `totalSupply` (the fact that this isn't done _inside_ the function is why I label it "unsafe"). As long as `totalSupply` fits in 112 bits (after the increment), each user's balance will also fit in 112 bits. Therefore, we can get away with unchecked math here.

See `///@dev` comment at the top of the function

core/src/Lender.sol
IanWoodard: I understand the reasoning, but it would save gas to combine them.

core/src/Lender.sol
IanWoodard: Oh, I didn't see that this was in an unchecked already, my bad.

core/src/Lender.sol
```diff
@@ -330,22 +322,89 @@ contract Lender is Ledger {
...
+            // From most to least significant...
+            // -------------------------------
+            // | courier id       | 32 bits  |
+            // | user's principle | 112 bits |
+            // | user's balance   | 112 bits |
+            // -------------------------------
+            uint256 data = balances[to];
+
+            if (data >> 224 != 0) {
+                // Keep track of principle iff courier deserves credit
```
IanWoodard, line 359: What do you mean by "deserves credit"?

IanWoodard (approved):

PR #10, opened 2023-01-01
Move Borrower positions read/write logic into its own library

haydenshively:
This makes Borrower.sol cleaner, at the expense of a few dozen gas. Is the abstraction worth it, especially given that we only use each function (read, write) once?


core/src/libraries/Positions.sol
```diff
@@ -0,0 +1,75 @@
...
+            stor[1] = mem[1];
+        }
+        if (count > 2) {
+            require(mem[2] != mem[0] || mem[3] != mem[1]);
+            stor[2] = mem[2];
+            stor[3] = mem[3];
+        }
+        if (count > 4) {
+            require(mem[4] != mem[0] || mem[5] != mem[1]);
+            require(mem[4] != mem[2] || mem[5] != mem[3]);
```
IanWoodard: Why separate these?

core/src/libraries/Positions.sol
```diff
@@ -0,0 +1,75 @@
...
+            // cpy:
+            // -->  xl + (xu << 24) + (yl << 48) + (yu << 72) + (zl << 96) + (zu << 120)
+            // -->  |-------|-----|----|----|----|----|----|
+            //      | shift | 120 | 96 | 72 | 48 | 24 |  0 |
+            //      | value |  zu | zl | yu | yl | xu | xl |
+            //      |-------|-----|----|----|----|----|----|
+            let cpy := sload(positions.slot)
+
+            ptr := mload(0x40)
+            let offset := 3
```
haydenshively: This should say := 32. Fixing in next commit, with tests

core/src/libraries/Positions.sol
```diff
@@ -0,0 +1,86 @@
...
+            stor[5] = 0;
+        } else if (count == 4) {
+            require(mem[0] != mem[2] || mem[1] != mem[3]);
+            stor[0] = mem[0];
+            stor[1] = mem[1];
+            stor[2] = mem[2];
+            stor[3] = mem[3];
+            stor[4] = 0;
+            stor[5] = 0;
+        } else {
```
IanWoodard: Why not make this an else if and remove the initial `count == 0` guard clause? It removes one check done every call in return for one that is only sometimes checked.

IanWoodard (approved):
Overall, LGTM. Left a few comments about some smaller optimizations.

core/src/libraries/Positions.sol
IanWoodard: Since there is no additional logic outside of the if statements

core/src/libraries/Positions.sol
haydenshively: Combined ✅

core/src/libraries/Positions.sol
haydenshively: Yeah I was going back and forth on this. It saves runtime gas in the `count == 0` case obviously, but each additional conditional branch adds substantial deploy cost, so we're probably better off with the setup you suggest.

PR #12, opened 2023-01-04
Add ETH ante to reward liquidators even for small accounts

haydenshively:
This is still a draft because I need to make the logic more modular, add comments, ~~add pure ETH gas compensation~~, and write tests.


core/src/Borrower.sol
```diff
@@ -52,7 +60,7 @@ contract Borrower is IUniswapV3MintCallback {
         bool isInCallback;
     }
 
-    struct SolvencyCache {
+    struct Prices {
         uint160 a;
         uint160 b;
         uint160 c;
```
IanWoodard: What does a, b, and c stand for?

IanWoodard: Can we get some comments?

core/src/Borrower.sol
```diff
@@ -79,27 +87,83 @@ contract Borrower is IUniswapV3MintCallback {
...
+        }
+
+        uint256 priceX96 = Math.mulDiv(prices.c, prices.c, FixedPoint96.Q96);
+
+        if (liabilities0 > 0) {
+            TOKEN1.safeApprove(address(callee), type(uint256).max);
+            callee.callback0(data, assets1, liabilities0);
+            TOKEN1.safeApprove(address(callee), 0);
+
+            uint256 repay0 = liabilities0 - LENDER0.borrowBalanceStored(address(this));
```
IanWoodard: What is borrowBalanceStored?

core/src/Borrower.sol
```diff
@@ -79,27 +87,83 @@ contract Borrower is IUniswapV3MintCallback {
...
+            uint256 maxLoss1 = Math.mulDiv(repay0 * 1.05e9, priceX96, 1e9 * FixedPoint96.Q96);
```
IanWoodard: Why is it 1.05, and why isn't it a constant?

IanWoodard: I realize it may be for gas savings, but a comment explaining what 1.05 means would be useful.

core/src/Borrower.sol
```diff
@@ -79,27 +87,83 @@ contract Borrower is IUniswapV3MintCallback {
...
+            uint256 maxLoss0 = Math.mulDiv(repay1 * 1.05e9, FixedPoint96.Q96, 1e9 * priceX96);
+            require(assets0 - TOKEN0.balanceOf(address(this)) <= maxLoss0);
+
+            // TODO is scaling the best incentive, or would it be better to use thresholding?
+            payable(msg.sender).transfer((ANTE * repay1) / liabilities1);
+        }
     }
 
-    function modify(IManager callee, bytes calldata data, bool[2] calldata allowances) external {
+    function modify(IManager callee, bytes calldata data, bool[2] calldata allowances) external payable {
```
IanWoodard, line 145: what does the payable mean in this context?

core/src/Borrower.sol
```diff
@@ -113,22 +177,26 @@ contract Borrower is IUniswapV3MintCallback {
...
-        require(
-            _isSolvent(
-                positions_,
-                Uniswap.FeeComputationCache(
-                    currentTick,
-                    UNISWAP_POOL.feeGrowthGlobal0X128(),
-                    UNISWAP_POOL.feeGrowthGlobal1X128()
-                ),
-                _getSolvencyCache()
+        Prices memory c = _getPrices();
```
IanWoodard: Is this different from the other c used in this context (in the price array)? If so, it is pretty confusing to have c contain a different c.

core/src/Borrower.sol
```diff
@@ -113,22 +177,26 @@ contract Borrower is IUniswapV3MintCallback {
...
+                UNISWAP_POOL.feeGrowthGlobal0X128(),
+                UNISWAP_POOL.feeGrowthGlobal1X128()
             ),
-            "Aloe: need more margin"
+            c
         );
+        (uint256 liabilities0, uint256 liabilities1) = _getLiabilities();
+        require(_isSolvent(liabilities0, liabilities1, assets, c), "Aloe: need more margin");
+
+        if (liabilities0 + liabilities1 > 0) require(address(this).balance > ANTE, "Aloe: missing ante");
```
IanWoodard, line 170: Could you explain the meaning of this line?

core/src/Borrower.sol
```diff
@@ -191,38 +259,86 @@ contract Borrower is IUniswapV3MintCallback {
         if (amount1 != 0) TOKEN1.safeTransfer(msg.sender, amount1);
     }
 
+    function _getAssetsWithdrawingPositions(Prices memory c2) private returns (Assets memory assets) {
```
IanWoodard: same question as above, c2 is very confusing

core/src/Borrower.sol
```diff
@@ -191,38 +259,86 @@ contract Borrower is IUniswapV3MintCallback {
...
+        assets.fixed0 = TOKEN0.balanceOf(address(this));
+        assets.fixed1 = TOKEN1.balanceOf(address(this));
+
+        int24[] memory positions_ = positions.read();
+        uint256 count = positions_.length;
+        delete positions;
```
IanWoodard: Why delete positions? I get this is probably some kind of gas optimization, but I am not sure where positions come into it (rather than positions_)

core/src/Borrower.sol
```diff
@@ -191,38 +259,86 @@ contract Borrower is IUniswapV3MintCallback {
...
+                    (uint256 burned0, uint256 burned1) = UNISWAP_POOL.burn(l, u, liquidity);
+                    (uint256 collected0, uint256 collected1) = UNISWAP_POOL.collect(
+                        address(this),
+                        l,
+                        u,
+                        type(uint128).max,
+                        type(uint128).max
+                    );
+                    assets.fixed0 += collected0 - burned0;
+                    assets.fixed1 += collected1 - burned1;
```
IanWoodard: Why are we subtracting burned?

core/src/Borrower.sol
```diff
@@ -191,38 +259,86 @@ contract Borrower is IUniswapV3MintCallback {
...
+
+                    // Compute lower and upper sqrt ratios
+                    L = TickMath.getSqrtRatioAtTick(l);
+                    U = TickMath.getSqrtRatioAtTick(u);
+                }
+
+                // Compute the value of `liquidity` (in terms of token1) at both probe prices
+                assets.fluid1A += LiquidityAmounts.getValueOfLiquidity(c2.a, L, U, liquidity);
+                assets.fluid1B += LiquidityAmounts.getValueOfLiquidity(c2.b, L, U, liquidity);
```
IanWoodard: Are we assuming fluids are always liquid in this case?

core/src/Borrower.sol
```diff
@@ -191,38 +259,86 @@ contract Borrower is IUniswapV3MintCallback {
...
+    ) private pure returns (bool) {
         // liquidation incentive. counted as liability because account will owe it to someone.
         // compensates liquidators for inventory risk.
         uint256 liquidationIncentive = _computeLiquidationIncentive(
             mem.fixed0 + mem.fluid0C,
             mem.fixed1 + mem.fluid1C,
             liabilities0,
             liabilities1,
-            c2.c
+            prices.c
```
IanWoodard: much better naming!

IanWoodard (commented):
Good start! Left some initial comments

core/src/Borrower.sol
```diff
@@ -79,27 +87,83 @@ contract Borrower is IUniswapV3MintCallback {
...
+            TOKEN1.safeApprove(address(callee), type(uint256).max);
+            callee.callback0(data, assets1, liabilities0);
+            TOKEN1.safeApprove(address(callee), 0);
+
+            uint256 repay0 = liabilities0 - LENDER0.borrowBalanceStored(address(this));
+            uint256 maxLoss1 = Math.mulDiv(repay0 * 1.05e9, priceX96, 1e9 * FixedPoint96.Q96);
+            require(assets1 - TOKEN1.balanceOf(address(this)) <= maxLoss1);
 
-        if (!isSolvent) packedSlot.owner = msg.sender;
+            // TODO is scaling the best incentive, or would it be better to use thresholding?
```
IanWoodard, line 124: Can you expand on what you mean by scaling vs thresholding?

core/src/Borrower.sol
haydenshively: It means that users can send ETH along with a function call to `modify`

core/src/Borrower.sol
haydenshively: yes, in the process of renaming this to `prices`

core/src/Borrower.sol
haydenshively: 1.05 implies that the liquidator receives a 5% reward. I will make a constant (constants don't impact gas)

PR #13, opened 2023-01-11
Add deploy scripts and use RPC endpoints properly

haydenshively:


core/script/Deploy.s.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import "forge-std/Script.sol";
+
+import {Borrower} from "../src/Borrower.sol";
```
aneesh1: Unused import?

core/script/Deploy.s.sol
```diff
@@ -0,0 +1,24 @@
...
+import {Factory, IUniswapV3Pool} from "../src/Factory.sol";
+import {Lender} from "../src/Lender.sol";
```
aneesh1: Unused import - not sure if this needed.

periphery/foundry.toml
```diff
@@ -7,6 +7,7 @@ libs = ['lib']
 out = '../build/periphery'
 cache_path = '../cache/periphery'
 
+via_ir = true
```
aneesh1, line 10: I thought we were shying away from this - why are we using this now?

aneesh1 (approved):
2 nits and 1 concern, otherwise good.

periphery/foundry.toml
haydenshively: After discussing it with you, I went and actually tried it out for the first time in a while. My primary concern was that it was still unstable, but I read through recent Solidity changelogs and it seems like in newer versions, there are actually _more_ bugs with the _old_ compilation pipeline than with the IR one. Plus it saved more gas than I expected.

So going forward, my thoughts are:
- Do our best to make things compile with `via_ir = false`, since it's nice to be able to go that route for quick testing (it compiles much much faster)
- Use `via_ir = true` in prod

IanWoodard: We were, but it has been updated quite a bit and we tested it again recently and found it helps us optimize a bit.

aneesh1: Good to know

PR #15, opened 2023-01-11
Remove logic that would've increased likelihood of overpaying in repay

haydenshively:
Before this change, the logic in `lender.repay()` allowed the caller to specify a repay amount _greater_ than their total liabilities. If they did (or if they simply passed 0) the logic would compute the maximum repay amount, and use that. In other words, `actualRepayAmount = min(specifiedRepayAmount, allDebt)`. This is great, except for the fact that the user may accidentally transfer too much money into the Lender before the `repay()` call.

Usually, repayment looks something like this:
```Solidity
lender.asset().safeTransfer(address(lender), specifiedRepayAmount);
lender.repay(specifiedRepayAmount, address(this));
```

So the problem was that they've already transferred `specifiedRepayAmount`, only for it to possibly get clamped by the `min(...)` thing described above.

This change prevents that from happening. If they try to repay too much, the transaction will just revert. That way, it's easier to avoid transferring in too many assets.


core/src/Lender.sol
```diff
@@ -187,10 +187,9 @@ contract Lender is Ledger {
         (Cache memory cache, ) = _load();
 
         unchecked {
-            if (amount == 0 || (units = (amount * BORROWS_SCALER) / cache.borrowIndex) >= b) {
-                units = b - 1;
-                amount = units.mulDivUp(cache.borrowIndex, BORROWS_SCALER);
-            }
+            units = (amount * BORROWS_SCALER) / cache.borrowIndex;
+            require(units < b, "Aloe: repay too much");
```
aneesh1, line 191: Note: if line 190 covers this comment, feel free to ignore. 

My main concern is that this is too restrictive. Suppose a borrower makes a small math mistake in their computations. Their transaction will be reverted, but the bigger problem is that in the next block, they might accrue more debt. I agree that overpaying massive amounts should be something we defend against, but perhaps some overpaying threshold to account for minor errors. How often do minor math mistakes happen in solidity? If not often, my point is moot so ignore. If often, then I think some threshold would be better.

aneesh1 (changes requested):

core/src/Lender.sol
haydenshively: Well if they're doing everything on chain (like from another contract), full repays can be done perfectly:
```
uint256 repayAmount = lender.borrowBalance(user);
lender.asset().safeTransfer(address(lender), repayAmount);
lender.repay(repayAmount, user);
```
I made sure that we round in the right direction when dividing, so that the above code works every time.

If they're doing things off-chain, and compute that they owe 10 tokens at timeA, but their transaction doesn't get included until timeB, then you're right: they won't have repaid their full debt. But for reasonable timeframes (tens of seconds), that discrepancy is small enough to ignore. And it won't revert because they're repaying _less_ than the max amount.

IanWoodard (approved):
LGTM

core/src/Lender.sol
aneesh1: I think I'm talking about something slightly different. My concern was a case like this: borrower owes 10 tokens at timeA. They then repay 10.2 tokens for example. I understand that repays can be done perfectly, but what I was concerned about is that in this change, the transaction would be reverted and the borrower could go into more debt. If I'm understanding correctly, between two blocks, they won't go into THAT much more debt so it seems reasonable to expect perfect repays each time.

aneesh1 (approved):

core/src/Lender.sol
haydenshively: Why would they repay 10.2 tokens if they only owe 10?

haydenshively: I think I see what you mean though; it'd be nice to be able to express the intent "just pay off all my debt" without worrying about exact computation. In fact, that's what the special `0` case was for previously -- it would magically pay off all your debt.

Honestly there are tradeoffs both ways. Old way you may accidentally overpay, new way it's hard to be exact, so you may revert or be left with some "dust" debt.

The real solution is to use a pull architecture (`erc20.safeTransferFrom(user, lender, exactAmountNeeded)`) instead of the push architecture (where we expect the user to have already `erc20.safeTransfer`ed). But right now, the rest of the contract uses push, and I want to stay consistent. Technically ERC4626 requires us to implement pull in deposit though, so if I decide to do that, maybe I can add it here too.

Sorry for the long msg!

PR #17, opened 2023-01-11
Bring periphery contracts in line with core

haydenshively:


periphery/src/BorrowerLens.sol
```diff
@@ -1,19 +1,22 @@
...
+import {Borrower, IUniswapV3Pool} from "aloe-ii-core/Borrower.sol";
 
 contract BorrowerLens {
     using Uniswap for Uniswap.Position;
 
     function getLiabilities(Borrower account) external view returns (uint256 amount0, uint256 amount1) {
-        amount0 = Lender(account.LENDER0()).borrowBalance(address(account));
-        amount1 = Lender(account.LENDER1()).borrowBalance(address(account));
+        amount0 = account.LENDER0().borrowBalance(address(account));
+        amount1 = account.LENDER1().borrowBalance(address(account));
```
aneesh1, line 13: `borrowBalance` or `borrowBalanceStored`? Former accounts for interest, the latter doesn't. Solvency check in borrower [uses `borrowBalanceStored`](https://github.com/aloelabs/aloe-ii/blob/8d0111d2ff497e60e5a25b551e092428edc41710/core/src/Borrower.sol#L300-L303). Is it ok that the two don't use the same function?

periphery/src/LenderLens.sol
```diff
@@ -22,10 +22,11 @@ contract LenderLens {
...
     {
         asset = lender.asset();
-
         (totalSupply, inventory, totalBorrows) = lender.stats();
 
-        if (inventory != 0) utilization = Math.mulDiv(1e18, totalBorrows, inventory);
-        interestRate = utilization > 0.5e18 ? 1.24e9 : 6.27e8;
+        if (inventory != 0) {
+            utilization = Math.mulDiv(1e18, totalBorrows, inventory);
+            interestRate = lender.rateModel().computeYieldPerSecond(utilization);
```
aneesh1, line 29: neat

aneesh1 (approved):
good

periphery/src/BorrowerLens.sol
haydenshively: This is a good callout. In general, `borrowBalance` is the more "correct" one because (as you said) it includes interest _up to and including the current block._ The reason I use `borrowBalanceStored` in the solvency check is that it's slightly more gas-efficient, and if that one additional chunk of interest-accrual is what pushes a user over the edge, liquidators can call `accrueInterest()` themselves. After calling `accrueInterest`, `borrowBalance == borrowBalanceStored`.

Here, since it's a Lens, the primary use-case is frontends fetching data -- so we don't have to worry about gas, and I can use the always-accurate `borrowBalance`

PR #18, opened 2023-01-11
Read positions from storage if callback doesn't pass any

haydenshively:
Fixes a bug with the current positions logic.

With the current code, the expectation is that the `IManager` will return an empty array if they don't wish to change/remove any of their Uniswap Positions. In that case, `Positions.write` doesn't write anything to storage, which is correct. However, it's reasonable to assume that the `IManager` still wants their Uniswap positions to be included as assets during the solvency check.

Before this fix: `positions_` is still an empty array when we start computing assets. If `IManager` passes a non-empty array, then those positions _will_ be included as assets, but they'll also be written to storage, which is wasteful.

**Goal:** Allow `IManager` to express the intent "I don't want to change my positions, but I do want my existing positions to be included as assets"

After this fix: If `IManager` passes an empty array, `Positions.write` will return existing positions. Otherwise it returns its input.


core/.gas-snapshot
```diff
@@ -1,15 +1,15 @@
 BorrowerGasTest:test_addMargin() (gas: 16075)
-BorrowerGasTest:test_borrow() (gas: 97662)
+BorrowerGasTest:test_borrow() (gas: 100023)
```
haydenshively, line 2: Many of these increased by ~2000 gas, because they represent the case where `IManager.callback` is returning an  empty array -- before we did nothing, now we have to read from storage. `sload` costs 2000 gas.

IanWoodard (approved):
The logic makes sense, but something I want more clarity on: so if the user has, say 3 positions and passes in a single position to write, all three of the positions are overwritten, and the new position is added, right? But if the user passes an empty array, everything is untouched? Would the user have to pass in [0, 0] in mem to clear positions? Also, isn't it possible with the current implementation to have negative Uniswap positions?

haydenshively:
> The logic makes sense, but something I want more clarity on: so if the user has, say 3 positions and passes in a single position to write, all three of the positions are overwritten, and the new position is added, right? But if the user passes an empty array, everything is untouched? Would the user have to pass in [0, 0] in mem to clear positions? Also, isn't it possible with the current implementation to have negative Uniswap positions?

Correct, correct, correct. ~Only thing that's maybe not correct: what do you mean by negative Uniswap positions?~ Discussed offline, good to go.

PR #19, opened 2023-01-11
Isolate positions extraction logic in preparation for rewrite

haydenshively:


IanWoodard (approved):
This looks good but could you give more context as to why the change, you left the description empty.

haydenshively:
It's because I need to use the `extract` function in multiple places in the next PR. Just breaking down the changes so it's easier to review

PR #20, opened 2023-01-11
Simplify IManager callback API with zip() function

haydenshively:
This API is both simpler (send 0 if you don't want to update, send new list of positions if you do) and more gas efficient.

The only non-intuitive thing is that if your intent is "no longer consider _any_ positions" you _cannot_ do `return zip([0, 0, 0, 0, 0, 0])`, as this is the same as `return 0`, which tells the code not to write to storage.

Instead, you have to do something like `return zip([0xDEAD, 0xDEAD, 0, 0, 0, 0])`. We will need to call this out in the documentation, but it's actually a good thing. Next time the person goes to add positions, it'll be cheaper to edit storage from {non-zero value} to {another non-zero value} than it is to go from {0} to {non-zero value}

* Note: It doesn't have to be `0xDEAD`, it can be any int24. You can even return non-zero values in the other positions if you want, as long as `firstPartOfEachPair == secondPartOfEachPair`. In other words, `return zip([1, 1, 2, 2, 3, 3])` would clear out your positions just as well as the 0xDEAD thing


core/src/Borrower.sol
```diff
@@ -107,15 +107,12 @@ contract Borrower is IUniswapV3MintCallback {
         if (allowances[1]) TOKEN1.safeApprove(address(callee), type(uint256).max);
 
         packedSlot.isInCallback = true;
-        int24[] memory positions_ = callee.callback(data);
+        int24[] memory positions_ = positions.write(callee.callback(data));
```
IanWoodard, line 110: I was curious about why this was separated in the earlier PR but this makes sense

IanWoodard: Is there anything that happens between these two calls that could affect this?

core/src/libraries/Positions.sol
```diff
@@ -14,7 +30,7 @@ uint256 constant Q24 = 0x1000000;
  * @param zipped Encoded Uniswap positions. Equivalent to the layout of `int24[6] storage yourPositions`
  * @return positionsOfNonZeroWidth Flattened array of Uniswap positions that may or may not hold liquidity
  */
-function extract(uint144 zipped) pure returns (int24[] memory positionsOfNonZeroWidth) {
+function extract(uint256 zipped) pure returns (int24[] memory positionsOfNonZeroWidth) {
```
IanWoodard, line 33: Why the change from uint144?

core/src/libraries/Positions.sol
```diff
@@ -59,44 +75,37 @@ function extract(uint144 zipped) pure returns (int24[] memory positionsOfNonZero
 }
 
 library Positions {
-    function write(int24[6] storage stor, int24[] memory mem) internal returns (int24[] memory) {
-        // Validate formatting of Uniswap positions
-        uint256 count = mem.length;
+    function write(int24[6] storage positions, uint256 update) internal returns (int24[] memory) {
```
IanWoodard, line 78: fan of the naming

core/src/libraries/Positions.sol
IanWoodard: I assume this is because it is stored as uint256 by Ethereum on the blockchain.

IanWoodard (approved):
Overall, LGTM! Good work!

core/src/libraries/Positions.sol
haydenshively: Yeah; it's because under the hood everything is in 256 bit chunks, and Yul (the language used in in-line assembly blocks) treats everything as uint256.

core/src/Borrower.sol
haydenshively: It's just because `write` didn't return anything before, so we had to store the callbacks' return value separately. Code in-between is unrelated.

PR #21, opened 2023-01-11
Delete vestigial periphery contracts and update for IManager API

haydenshively:


periphery/src/FrontendManager.sol
```diff
@@ -23,14 +23,14 @@ contract FrontendManager is IManager {
...
-    function callback(bytes calldata data) external returns (int24[] memory positions) {
+    function callback(bytes calldata data) external returns (uint144 positions) {
         require(FACTORY.isBorrower(msg.sender), "Aloe: bad account");
 
         Borrower account = Borrower(msg.sender);
 
         uint8[] memory actions;
         bytes[] memory args;
-        (actions, args, positions) = abi.decode(data, (uint8[], bytes[], int24[]));
+        (actions, args, positions) = abi.decode(data, (uint8[], bytes[], uint144));
```
IanWoodard, line 33: Is this because the value is zipped?

IanWoodard (approved):

periphery/src/FrontendManager.sol
haydenshively: Yep!

PR #23, opened 2023-01-12
Make liquidate() more conservative w.r.t. liquidator rewards

haydenshively:
This includes the main logic for upgraded liquidations. I've left out the ante for now, and there are still various TODOs. I'm especially bothered by the fact that we have 2 `_getAssets` methods; one that's view-only and one that withdraws positions. But I want to let you guys review this first.


core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
-            Uniswap.FeeComputationCache(
-                currentTick,
-                UNISWAP_POOL.feeGrowthGlobal0X128(),
-                UNISWAP_POOL.feeGrowthGlobal1X128()
-            ),
-            prices
-        );
+        Assets memory assets = _getAssetsAndWithdrawPositions(positions.read(), prices);
+
+        require(!BalanceSheet.isHealthy(liabilities0, liabilities1, assets, prices));
```
haydenshively, line 94: We start by checking that the account's balance sheet is unhealthy, i.e. that it can actually be liquidated. In the process of summing up assets, we've also withdrawn all Uniswap positions (so that money is now sitting in this contract, rather than Uniswap)

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
+
+        uint256 assets0 = TOKEN0.balanceOf(address(this));
+        uint256 repayable0 = Math.min(assets0, liabilities0);
+        liabilities0 -= repayable0;
+
+        uint256 assets1 = TOKEN1.balanceOf(address(this));
+        uint256 repayable1 = Math.min(assets1, liabilities1);
+        liabilities1 -= repayable1;
```
haydenshively, line 102: Next, we figure out how much of liabilities can be repaid using assets that are already sitting in the contract. No swaps are necessary to repay this debt, so we don't need the liquidator to do much other than call this function.

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
+        // TODO: this is already computed inside of computeLiquidationIncentive, so use that instead of
+        // re-computing it (or at least compare gas between the 2 options)
+        uint256 priceX96 = Math.mulDiv(prices.c, prices.c, FixedPoint96.Q96);
 
-        if (!BalanceSheet.isHealthy(liabilities0, liabilities1, assets, prices)) packedSlot.owner = msg.sender;
+        if (liabilities0 + liabilities1 == 0 || (liabilities0 > 0 && liabilities1 > 0)) {
+            // If both are zero or neither is zero, there's nothing more to do
+            // TODO: compensate liquidators for txn costs using ANTE
+            return;
+        } else if (liabilities0 > 0) {
```
haydenshively, line 112: For any remaining liabilities, we _do_ need the liquidator to get more involved. They have to swap the asset that's in surplus for the asset that's lacking.

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
+            TOKEN1.safeApprove(address(callee), type(uint256).max);
+            uint256 converted0 = callee.callback0(data, assets1 - repayable1, liabilities0);
+            TOKEN1.safeApprove(address(callee), 0);
+
+            uint256 maxLoss1 = Math.mulDiv(converted0 * 105, priceX96, 100 * FixedPoint96.Q96);
+            require(assets1 - TOKEN1.balanceOf(address(this)) <= maxLoss1);
```
haydenshively, line 118: Inside the callback, the `callee` was expected to take up to `assets1` worth of token1, and swap it to generate `converted0` worth of token0. The `callee`'s reward is based on how much they _actually_ swapped.

Basically, if the price of ETH is $1000 and `converted0` is 1000 USDC, we'd allow `callee` to take up to 1.05 ETH -- enough to execute a slippage-less swap, plus 5% reward.

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
+
+            // TODO: compensate liquidators for txn costs using ANTE
+
+            repayable0 += converted0;
+        } else {
+            TOKEN0.safeApprove(address(callee), type(uint256).max);
+            uint256 converted1 = callee.callback1(data, assets0 - repayable0, liabilities1);
+            TOKEN1.safeApprove(address(callee), 0);
```
IanWoodard: Shouldn't it be TOKEN0 for which we are revoking approval here?

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
+
+            uint256 maxLoss0 = Math.mulDiv(converted1 * 105, FixedPoint96.Q96, 100 * priceX96);
```
IanWoodard, line 128: Should the divisor be FixedPoint96.Q96 * 100?

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
         packedSlot.owner = owner;
     }
 
-    // TODO liquidations
-    function liquidate() external {
+    function liquidate(ILiquidator callee, bytes calldata data) external {
         require(!packedSlot.isInCallback);
```
IanWoodard, line 89: Is there any protection against a reentrancy attack?

IanWoodard (commented):
Left some initial comments

core/src/Borrower.sol
haydenshively: Oof, yes good catch

core/src/Borrower.sol
haydenshively: No, that goes in the numerator here.

core/src/Borrower.sol
haydenshively: Well they can't reenter from modify's callback, because of the require(!isInCallback). But they could reenter from an ILiquidator contract. Will need to think more about whether that's dangerous or not.

haydenshively: Thought about this some more, and as far as this PR goes, I think it's safe without a reentrancy check. But once I add the ante back in, an attacker could take advantage of reentrancy to (potentially) steal all the ante when they only deserve some portion of it.

core/src/Borrower.sol
```diff
@@ -185,6 +229,57 @@ contract Borrower is IUniswapV3MintCallback {
...
+                    int24 l = positions_[i];
+                    int24 u = positions_[i + 1];
+                    // Fetch amount of `liquidity` in the position
+                    (liquidity, , , , ) = UNISWAP_POOL.positions(keccak256(abi.encodePacked(address(this), l, u)));
+
+                    // Compute lower and upper sqrt ratios
+                    L = TickMath.getSqrtRatioAtTick(l);
+                    U = TickMath.getSqrtRatioAtTick(u);
+
+                    // TODO what happens if we attempt to withdraw a position that doesn't exist?
```
aneesh1: Following the calls:
1. `burn` calls [`_modifyPosition`](https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L528-L535)
2. `_modifyPostion` calls [`_updatePosition`](https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L322-L328)
3. `_updatePosition` calls [`positions.get`](https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L389) - under the hood this is essential just querying the `mapping`. If my understanding of `mapping` is correct, this should succeed and will return [default values (all zeroes)](https://docs.soliditylang.org/en/v0.4.21/types.html#mappings) for all fields of the `Position`. 
4. We then call [`position.update`](https://github.com/Uniswap/v3-core/blob/05c10bf6d547d6121622ac51c457f93775e1df09/contracts/UniswapV3Pool.sol#L449) using the `position` object returned in step 3. It'll revert [here](https://github.com/Uniswap/v3-core/blob/05c10bf6d547d6121622ac51c457f93775e1df09/contracts/libraries/Position.sol#L54).

aneesh1 (commented):
Overall looks good! See comment about attempting to withdraw a position that doesn't exist.

core/src/Borrower.sol
haydenshively: Wrote a quick test to confirm this, and you're 100% correct. Added a conditional to `continue` early if `liquidity == 0`.

core/src/Borrower.sol
haydenshively: Made a note of this in #25

core/src/Borrower.sol
```diff
@@ -70,26 +84,56 @@ contract Borrower is IUniswapV3MintCallback {
...
+
+        // TODO: this is already computed inside of computeLiquidationIncentive, so use that instead of
+        // re-computing it (or at least compare gas between the 2 options)
+        uint256 priceX96 = Math.mulDiv(prices.c, prices.c, FixedPoint96.Q96);
 
-        if (!BalanceSheet.isHealthy(liabilities0, liabilities1, assets, prices)) packedSlot.owner = msg.sender;
+        if (liabilities0 + liabilities1 == 0 || (liabilities0 > 0 && liabilities1 > 0)) {
+            // If both are zero or neither is zero, there's nothing more to do
+            // TODO: compensate liquidators for txn costs using ANTE
+            return;
```
haydenshively, line 111: Should fall through to line 136, not `return`. Addressed in #27

aneesh1 (approved):
Good call making #25 - comments addressed, this looks good.

core/src/Borrower.sol
IanWoodard: ~~Wouldn't this withdraw the positions regardless of whether the balance sheet is healthy or not? If so, wouldn't that pose problems since bad actors could withdraw users' positions even if they are solvent? With that said, if there is something I am missing, please enlighten me.~~ Aneesh cleared this up for me; I forgot about the whole revert unwounds everything idea.

IanWoodard (approved):

PR #26, opened 2023-01-14
Consolidate Borrower's two getAssets functions; move fee-views to periphery

haydenshively:
More than just consolidating the two getAssets function, this does actually change behavior:

Prior to this change, the version of getAssets used in both `liquidate` and `modify` included uncollected Uniswap fees as assets. For `liquidate`, we're burning all Uniswap positions anyway, so fees are naturally included. But in `modify` we had to be fancy and fetch some extra state (current tick, fee growth globals, and 2 things from mappings) to calculate uncollected fees without modifying state.

After this change, `modify`'s estimation of assets won't include uncollected fees. imo this is better UX, because:
- if a margin account doesn't have Uniswap Positions, that extra state we were fetching was wasteful
- if a margin account does have Uniswap Positions, it's likely that they'll be moving them around / depositing / withdrawing on every call to `modify`, which means fees have been collected. Collected fees are still counted as assets.
- this way, we force users to be a bit more conservative -- in `modify` their account needs to be healthy _without_ uncollected fees, but `liquidate` counts all fees in their favor. It's a little bonus safety margin for them

Since people may still want info about uncollected fees (and we show it in the frontend) I moved the fancy state stuff to BorrowerLens


IanWoodard (approved):
Not a huge fan of calling the function getAssets and then having it withdraw positions as well (I know it is a flag, but still), but other than that, LGTM

aneesh1 (approved):

PR #27, opened 2023-01-14
Add tests for liquidation functionality

haydenshively:


core/src/Borrower.sol
```diff
@@ -124,7 +123,7 @@ contract Borrower is IUniswapV3MintCallback {
             uint256 converted1 = callee.callback1(data, assets0 - repayable0, liabilities1);
             TOKEN0.safeApprove(address(callee), 0);
 
-            uint256 maxLoss0 = Math.mulDiv(converted1 * 105, Q96, 100 * priceX96);
+            uint256 maxLoss0 = Math.mulDiv(converted1, Q96, priceX96);
```
haydenshively, line 116: Slipped through the cracks on #24

core/src/Borrower.sol
```diff
@@ -82,14 +74,14 @@ contract Borrower is IUniswapV3MintCallback {
         packedSlot.owner = owner;
     }
 
-    function liquidate(ILiquidator callee, bytes calldata data) external {
+    function liquidate(ILiquidator callee, bytes calldata data, uint256 strain) external {
```
aneesh1, line 77: what is the purpose of `strain`?

aneesh1 (approved):
Tests look really good. Only one question.

core/src/Borrower.sol
haydenshively: Good question (and I will write docstrings to answer it too) -- previously, the liquidation callee decided what percentage of remaining liabilities to obtain through swapping, and informed `Borrower` through the callback return values (`converted0` and `converted1`). `Borrower` then computed a reward. Now, since I've switched the order, `Borrower` needs to know ahead of time how much the callee intends to swap.

When `strain == 1`, the callee is expected to swap enough assets to cover 100% of debt. When `strain == 2` it's 50%, `strain == 3` is 33%, and so on. In almost every case, liquidators will use `strain = 1` because that maximizes their reward. But if for some reason the market is "strained" or the debt is huge, and a swap that large can't go through, they can opt to only cover a portion of the debt by setting higher strain values.

PR #30, opened 2023-01-18
Avoid re-computing liquidation incentive to prevent tick manipulation

haydenshively:
Previously, lines 145 and 157 re-computed the liquidation incentive based on the actual, instantaneous shortfall. This would have encouraged liquidators to manipulate the price to artificially increase the shortfall, and earn a higher reward.

Now, we use `BalanceSheet.computeLiquidationIncentive` to compute the liquidation incentive once and only once, based on the TWAP. Liquidator's expected swap size is still dependent on instantaneous price, but their reward won't change, so there's no incentive to manipulate things.


core/src/Borrower.sol
```diff
@@ -103,46 +102,50 @@ contract Borrower is IUniswapV3MintCallback {
...
-
         if (liabilities0 + liabilities1 == 0 || (liabilities0 > 0 && liabilities1 > 0)) {
             // If both are zero or neither is zero, there's nothing more to do
             // TODO: compensate liquidators for txn costs using ANTE
         } else if (liabilities0 > 0) {
             uint256 converted0 = liabilities0 / strain;
 
-            uint256 maxLoss1 = Math.mulDiv(converted0, priceX96, Q96);
-            maxLoss1 += maxLoss1 / LIQUIDATION_INCENTIVE;
+            uint256 maxLoss1 = Math.mulDiv(converted0, priceX96, Q96) + incentive1 / strain;
```
aneesh1, line 148: I'm not 100% on order of operations in solidity, but here are you trying to do $$\frac{Math.mulDiv(converted0, priceX96, Q96) + incentive1}{strain}$$? Or are you trying to $$Math.mulDiv(converted0, priceX96, Q96) + \frac{incentive1}{strain}$$? I feel like it would be the former because the `strain` determines how much the callee needs to swap in order to cover the liabilities and the liquidation incentive is a potential liability.

core/src/Borrower.sol
```diff
@@ -153,8 +156,7 @@ contract Borrower is IUniswapV3MintCallback {
         } else {
             uint256 converted1 = liabilities1 / strain;
 
-            uint256 maxLoss0 = Math.mulDiv(converted1, Q96, priceX96);
-            maxLoss0 += maxLoss0 / LIQUIDATION_INCENTIVE;
+            uint256 maxLoss0 = Math.mulDiv(converted1 + incentive1 / strain, Q96, priceX96);
```
aneesh1, line 159: Same as previous comment.

core/src/libraries/BalanceSheet.sol
```diff
@@ -84,16 +93,16 @@ library BalanceSheet {
...
-            uint256 meanPriceX96 = Math.mulDiv(sqrtMeanPriceX96, sqrtMeanPriceX96, Q96);
+            meanPriceX96 = uint224(Math.mulDiv(sqrtMeanPriceX96, sqrtMeanPriceX96, Q96));
 
             if (liabilities0 > assets0) {
                 // shortfall is the amount that cannot be directly repaid using Borrower assets at this price
                 uint256 shortfall = liabilities0 - assets0;
                 // to cover it, a liquidator may have to use their own assets, taking on inventory risk.
                 // to compensate them for this risk, they're allowed to seize some of the surplus asset.
-                incentive1 += Math.mulDiv(shortfall, meanPriceX96, Q96) / LIQUIDATION_INCENTIVE;
+                incentive1 += Math.mulDiv(shortfall / LIQUIDATION_INCENTIVE, meanPriceX96, Q96);
```
aneesh1, line 105: Equivalent expressions 👍

aneesh1 (commented):
Two instances of the same question, otherwise, looks good

core/src/Borrower.sol
haydenshively: So, that's all true -- only thing is that `converted0` already includes the division by `strain`, so here I'm doing the 2nd set of LaTeX.

core/src/libraries/BalanceSheet.sol
haydenshively: 🙃

core/src/Borrower.sol
haydenshively: Actually let me try doing it all at once. ~Def would improve readability and now that I think about it, it may save us 3 gas as well (one opcode improvement lol)~

```Solidity
  uint256 maxLoss1 = (Math.mulDiv(liabilities0, priceX96, Q96) + incentive1) / strain;

  TOKEN1.safeTransfer(address(callee), maxLoss1);
  callee.swap1For0(data, maxLoss1, liabilities0 / strain);
  
  repayable0 += liabilities0 / strain;
```

Here's the alternate version. Saved one stack var but added another division, so gas is a wash. Let me know which one you think is more readable and I'll go with it.

core/src/libraries/BalanceSheet.sol
haydenshively: There was an actual reason for this though. `shortfall` is _at most_ equal to `liabilities0`, and because of constraints in Lender, the maximum value of liabilities is 2^184-1. The maximum value of `meanPriceX96` is 2^224-1, and `Q96` is 2^96. The result of `Math.mulDiv` must fit in 256 bits, which means if `liabilities0 > 2**128` this could fail. By dividing by `LIQUIDATION_INCENTIVE` _before_ entering the `mulDiv` step, I'm giving us an extra `log2(20)=4.32` bits of wiggle room.

Is this completely esoteric? Yes. Will token amounts > 2**128 break other things first? Almost certainly. But who doesn't love a 4.32 bit improvement?

core/src/Borrower.sol
aneesh1: I understand. I think the alternate version is more readable b/c it's more obvious that we're dividing `strain` by all liabilities.

haydenshively: Curious if the variable naming change [here](https://github.com/aloelabs/aloe-ii/pull/40/files) changes your mind.

haydenshively: @aneesh1 I suggest we table this for now, as we'll be more than able to address it once this slew of PRs is merged. I don't really want to edit this one, since everything else builds on it.

PR #38, opened 2023-01-18
Add ETH ante to reward liquidators even for small accounts

haydenshively:


core/src/Borrower.sol
```diff
@@ -189,7 +195,10 @@ contract Borrower is IUniswapV3MintCallback {
         Assets memory assets = _getAssets(positions_, prices, false);
         (uint256 liabilities0, uint256 liabilities1) = _getLiabilities();
 
-        require(BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: need more margin");
+        require(BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: unhealthy");
+        unchecked {
+            if (liabilities0 + liabilities1 > 0) require(address(this).balance > ANTE, "Aloe: missing ante");
```
aneesh1, line 200: Implicaton: for an account to be liquidatable it must possess at least `ANTE` amount of funds in their wallet before liquidation is permitted. Makes sense.

core/test/Liquidator.t.sol
```diff
@@ -39,6 +39,8 @@ contract LiquidatorTest is Test, IManager, ILiquidator {
         deal(address(asset1), address(lender1), 10000e18); // WETH
         lender0.deposit(10000e18, address(12345));
         lender1.deposit(10000e18, address(12345));
+
+        deal(address(account), account.ANTE() + 1);
```
aneesh1, line 43: To satisfy `require(address(this).balance > ANTE, "Aloe: missing ante");` - nice!

aneesh1 (commented):
Good

aneesh1 (approved):

core/src/Borrower.sol
haydenshively: Wait unless I messed up, this is actually in the `modify` function -- so we're saying that before someone can borrow any funds, they must possess at least `ANTE`. That way liquidators are guaranteed some kind of reward, even when swaps aren't necessary.

PR #41, opened 2023-01-18
Break packedSlot into its components

haydenshively:


core/src/Borrower.sol
```diff
@@ -51,12 +51,9 @@ contract Borrower is IUniswapV3MintCallback {
...
     Lender public immutable LENDER1;
 
-    struct PackedSlot {
-        address owner;
-        bool isInCallback;
-    }
+    address owner;
 
-    PackedSlot public packedSlot;
+    bool isInCallback;
```
aneesh1, line 56: Idea: instead of storing this as a boolean, store this as a unit96. So instead of 
```solidity
bool isInCallback;
```
what if we do:
```solidity
uint96 isInCallback;
```
Reasoning: `address` is 20 bytes, `bool` is 1 byte, so the `Borrower` is taking up 21 bytes. If we do `uint96` instead, then that would be 12 bytes, 20 + 12 = 32 bytes, and our `Borrower` would be 32 byte aligned. Not sure what else will change because of using `uint96` instead of `bool`, but I'm pretty confident this is an easy gas optimization. 

Sources:
1. https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html
2. https://dev.to/javier123454321/solidity-gas-optimizations-pt-3-packing-structs-23f4

aneesh1 (commented):
Good - one potential gas optimization, let me know what you think.

core/src/Borrower.sol
haydenshively: Since the two variables are already taking up less than 32 bytes, we only need a single `sload` (800 gas) to pull from storage into memory. Expanding the flag to take up 96 bits won't change that, although it _might_ save us a bit shift (3 gas) depending on what the compiler is doing.

To address #25 I may change this to a numeric flag anyway, but I'll probably stick with a `uint8` because I plan to pack more values into this slot in the future.

haydenshively: Somewhat related, I should do some testing to make sure that the compiler is optimizing things the way I thought it was. I know that when _reading_ multiple things from a single slot, it only does 1 `sload`, but when writing them it may do multiple `sload`/`sstore` pairs instead of just one. This could impact both `Lender` and `Borrower`.

Will get back to you.

aneesh1: 3 gas is hardly anything - disregard my comment.

aneesh1 (approved):

core/src/Borrower.sol
haydenshively: Alright so I went on quite the side quest, and determined that the compiler is in fact behaving as I thought it was. Did find one potential improvement and described it in #42, but that's separate from whether or not it's ok to break up a struct into its components like this.

PR #45, opened 2023-01-20
Add re-entrancy protection for Borrower.liquidate

### LINKED ISSUE #25: Ensure `liquidate()` is properly protected from re-entrancy once we add ante

haydenshively:
Closes #25


aneesh1 (approved):
good - want to enumerate what users can and can't do now:
- `uniswapDeposit` can only be called from within the `modify` callback
- `borrow` can only be called from within the `modify` callback
- `repay` can only be called from within the `modify callback` - was a bit concerned about this but then I saw that `liquidate` calls `_repay`, so we're good on that
- `modify` can be called from anywhere - moves to state `InModifyCallback` before executing the callback
- `liquidate` can be called from anywhere - once called, it moves to `Locked` state, so nothing else can be called until the function completes execution and returns to the `Ready` state

PR #46, opened 2023-01-20
Require warn() before liquidate() to mitigate manipulation risk

haydenshively:


core/src/Borrower.sol
```diff
@@ -88,6 +91,23 @@ contract Borrower is IUniswapV3MintCallback {
                            MAIN ENTRY POINTS
     //////////////////////////////////////////////////////////////*/
 
+    function warn() external {
```
aneesh1, line 94: ~~Want to emit an event?
`event insolvencyWarning (address unhealthyBorrower, uint256 amount0ToRepay, uint256 amount1ToRepay)`~~

~~`emit insolvencyWarning()`~~

Disregard - just saw this is addressed by https://github.com/aloelabs/aloe-ii/pull/48.

aneesh1 (approved):
Good - prior to calling `account.liquidate()` now you have to call `account.warn()` - should be a nice feature for users.

core/src/Borrower.sol
haydenshively: Next PR :)

haydenshively:
> Good - prior to calling `account.liquidate()` now you have to call `account.warn()` - should be a nice feature for users.

Note that this is only true for the branches of `liquidate` that trigger a swap callback and a 5% bonus. For the simpler liquidation (where we're able to repay everything in-kind without swapping) no warning is required.

PR #48, opened 2023-01-20
Add Warn and Liquidate events to Borrower

haydenshively:


core/src/Borrower.sol
```diff
@@ -277,6 +293,7 @@ contract Borrower is IUniswapV3MintCallback {
         if (amount1 > 0) LENDER1.borrow(amount1, recipient);
     }
 
+    // TODO: change/reword this
```
IanWoodard, line 296: Instead of putting a TODO here, what if you just made a GitHub issue?

core/src/Borrower.sol
```diff
@@ -82,11 +95,38 @@ contract Borrower is IUniswapV3MintCallback {
...
+            Assets memory assets = _getAssets(positions.read(), prices, false);
+            // Fetch liabilities from lenders
+            (uint256 liabilities0, uint256 liabilities1) = _getLiabilities();
+            // Ensure only unhealthy accounts get warned
+            require(!BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: healthy");
+        }
+
+        unleashLiquidationTime = uint88(block.timestamp + LIQUIDATION_GRACE_PERIOD);
+
+        emit Warn();
```
aneesh1, line 118: Nice! What are your thoughts on adding information about amount of debt they need to may back in either token 0 or token 1 to the event? Could that be useful?

aneesh1 (approved):
Good

core/src/Borrower.sol
IanWoodard: I was just writing about this as well. I think it would be good to add information about the warning in the event.

haydenshively: It'd definitely be useful, but those amounts (exactly how much they need to repay to avoid liquidation) aren't well-defined. We could just emit what `liabilities0` and `liabilities1` are, but that's pretty easy for them to check themselves so I'm not sure it's worth the gas.

Very very long-term, it'd be cool to make the protocol more aware of when and how different borrowers are going to be liquidated, so that it can plan ahead of time to execute swaps with less price impact. I suspect those computations will require some off-chain math with ZK proofs to be feasible though.

IanWoodard (approved):
Besides 2 questions I brought up, LGTM

core/src/Borrower.sol
IanWoodard: What about just general information?

core/src/Borrower.sol
haydenshively: Don't worry, I'll do another round of TODO cleanups soon :)

core/src/Borrower.sol
haydenshively: Discussed offline. For now it's fine as-is, but if we ever do a dynamic liquidation grace period, it'd be good to emit the deadline in this event.

PR #49, opened 2023-01-21
Repack Borrower slot0 for performance gains

haydenshively:
The main goal of this change was to make all components of slot0 accessible to other contracts on-chain, without needing multiple getters or manually-defined getters. A few side effects:
- runtime gas costs improve by 0.1%-1%, depending on the function (not much, but it's something)
- reduced market creation cost by ~100000 gas (not what we're optimizing for, but that's actually a decent amount)


IanWoodard (approved):
IMO this is less readable, but I understand the performance gains so LGTM

PR #50, opened 2023-01-21
Don't include fees in `_getAssets` regardless of `withdraw`

haydenshively:
We want the assets observed by `warn` and `liquidate` to match. Previously the view-only version and the withdrawing-positions version of `_getAssets` differed: if `withdraw==true`, uncollected Uniswap fees were included as fixed assets. This change removes that logic.

Basically, uncollected Uniswap fees are never counted as assets. imo this is a reasonable tradeoff given that users now receive warnings of liquidations, so if those uncollected fees are _really_ what makes the difference, they have a change to collect them and regain solvency. In practice, uncollected fees are probably a relatively small chunk of overall assets, so this shouldn't matter much.


aneesh1 (approved):
The two functions now compute assets in the same way - nice. Essentially, the `withdraw` flag has a new meaning: withdraw positions into the borrower's account. Curious about one thing you said in your initial comment:
> so if those uncollected fees are really what makes the difference, they have a chance to collect them and regain solvency

What does this look like? Would they use this action? 
Screen Shot 2023-01-21 at 9 34 45 PM

haydenshively:
> What does this look like? Would they use this action?

Yep!

Issue #33, opened 2023-01-18
ERC4626 demands `approve`/`transferFrom` flow


haydenshively:
Could add into core, or put in periphery as a sort of helper. Would include the pull-architecture version of deposit and a fallback function. Optionally move `mint` and `withdraw` out of core as well.

PR #52, opened 2023-01-24
Add invariant tests for lender

haydenshively:
LenderHarness.sol and LenderInvariants.sol are massive tests. Given that I only had to make one change (that wasn't even in our core business logic) to get everything passing, I think we can be pretty confident that Lender works as intended.


core/src/Ledger.sol
```diff
@@ -165,7 +165,7 @@ contract Ledger {
 
         (Cache memory cache, , ) = _previewInterest(_getCache());
         unchecked {
-            return (b - 1).mulDivUp(cache.borrowIndex, BORROWS_SCALER);
+            return ((b - 1) * cache.borrowIndex) / BORROWS_SCALER;
```
haydenshively, line 168: See #55

core/src/Lender.sol
```diff
@@ -412,7 +415,6 @@ contract Lender is Ledger {
         emit Transfer(from, address(0), shares);
     }
 
-    /// @dev Note that if `RESERVE` ever gives credit to a courier, its principle won't be tracked properly.
```
haydenshively, line 415: No longer a concern, because we make sure `RESERVE` _cannot_ credit a courier. Wasn't much of a concern before since `RESERVE`'s abilities are up to us anyway, but it doesn't hurt to be safe here.

core/src/Lender.sol
```diff
@@ -424,7 +426,7 @@ contract Lender is Ledger {
         (cache, inventory, newTotalSupply) = _previewInterest(cache);
 
         // Update reserves (new `totalSupply` is only in memory, but `balanceOf` is updated in storage)
-        if (newTotalSupply != cache.totalSupply) {
+        if (newTotalSupply > cache.totalSupply) {
```
haydenshively, line 429: Just more readable imo

core/src/Ledger.sol
aneesh1: Equivalent expressions 👍 - can we also anticipate some performance win similar to: https://github.com/aloelabs/aloe-ii/pull/30#discussion_r1073239600?

core/test/Lender.t.sol
```diff
@@ -223,7 +221,7 @@ contract LenderTest is Test {
         lender.accrueInterest();
 
         assertEq(asset.balanceOf(jim), 10e6);
-        assertEq(lender.borrowBalance(jim), 10000058);
+        assertEq(lender.borrowBalance(jim), 10000057);
```
aneesh1, line 224: This looks to me like evidence of rounding down, the behavior we were hoping for in #55.

core/test/Utils.sol
```diff
@@ -14,3 +14,10 @@ function deploySingleLender(ERC20 asset, address treasury, RateModel rateModel)
     Lender(proxy).initialize(rateModel, 8);
     return Lender(proxy);
 }
+
+contract Router {
```
aneesh1, line 18: Interesting - so I've seen other protocols use routers for defining the routing logic for transactions like trades. I see here we're doing something similar transferring tokens from `msg.sender` to the lender and then making a deposit. However the behavior is pretty simple, why do we need a `Router` for this?

core/test/invariants/LenderInvariants.t.sol
```diff
@@ -0,0 +1,221 @@
...
+import {LenderHarness, BORROWS_SCALER} from "./LenderHarness.sol";
+
+contract LenderInvariantsTest is Test, InvariantTest {
+    ERC20 public asset;
+
+    Lender public lender;
+
+    LenderHarness public lenderHarness;
+
+    struct ThingsThatShouldntChange {
```
aneesh1, line 22: Good naming

aneesh1 (approved):

core/src/Ledger.sol
haydenshively: These actually aren't equivalent! Previously we were rounding up, now we're rounding down.

(`mulDivUp` and `mulDivDown` are different from `FullMath.mulDiv`. I know it's confusing)

core/test/Lender.t.sol
haydenshively: Yes! I included a temporary fix to address the problems described in #55.

core/test/Utils.sol
haydenshively: It's because we're using a push architecture; i.e. by the time `deposit` or `repay` is called, `Lender` expects to have already been paid (it doesn't attempt to pull tokens in from the sender via `erc20.transferFrom`). Therefore, we need a router to help send tokens to the `Lender` right before each deposit.

haydenshively: Technically we're supposed to support both flows; see #33

PR #62, opened 2023-01-30
Add create2 deployment scripts

haydenshively:


aneesh1 (approved):
Had to read [this](https://blog.openzeppelin.com/getting-the-most-out-of-create2/) to understand the benefits of `CREATE2`. From this change, we can expected “domain parking” meaning we will know where a transaction will be deployed in the future. *It seems like we can also refer to contracts that haven’t been deployed. Not sure what good this does us when we’ve already deployed the contracts. Does it help us or users when making changes?*

This seems to have advantages when onboarding new users to DeFi. Not to mention, it seems like we can also deploy new contracts to the same address. Lastly, another key detail of `CREATE2` is that we need to provide the original sender’s address when constructing the salt. This has implications for our future workflows because only one of us can deploy new contracts. However, it seems like other users can deploy, but they need one of our signatures. Finally, the constructor plays an important role in determining the deployment address, so we need to be careful with changes to them and should use initializers instead

PR #63, opened 2023-01-31
Add VolatilityOracle

haydenshively:
I added additional tests for the underlying `Volatility.sol` library, but haven't yet written tests for the full `VolatilityOracle.sol`

### DIFF
--- a/core/src/Borrower.sol
+++ b/core/src/Borrower.sol
@@ -11,11 +11,11 @@ import {LIQUIDATION_GRACE_PERIOD} from "./libraries/constants/Constants.sol";
 import {Q96} from "./libraries/constants/Q.sol";
 import {BalanceSheet, Assets, Prices} from "./libraries/BalanceSheet.sol";
 import {LiquidityAmounts} from "./libraries/LiquidityAmounts.sol";
-import {Oracle} from "./libraries/Oracle.sol";
 import {Positions} from "./libraries/Positions.sol";
 import {TickMath} from "./libraries/TickMath.sol";
 
 import {Lender} from "./Lender.sol";
+import {VolatilityOracle} from "./VolatilityOracle.sol";
 
 interface ILiquidator {
     receive() external payable;
@@ -43,6 +43,9 @@ contract Borrower is IUniswapV3MintCallback {
     uint8 public constant B = 3;
 
     uint256 public constant ANTE = 0.001 ether;
+
+    /// @notice The oracle to use for prices and implied volatility
+    VolatilityOracle public immutable ORACLE;
 
     /// @notice The Uniswap pair in which the vault will manage positions
     IUniswapV3Pool public immutable UNISWAP_POOL;
@@ -79,7 +82,8 @@ contract Borrower is IUniswapV3MintCallback {
                        CONSTRUCTOR & INITIALIZER
     //////////////////////////////////////////////////////////////*/
 
-    constructor(IUniswapV3Pool pool, Lender lender0, Lender lender1) {
+    constructor(VolatilityOracle oracle, IUniswapV3Pool pool, Lender lender0, Lender lender1) {
+        ORACLE = oracle;
         UNISWAP_POOL = pool;
         LENDER0 = lender0;
         LENDER1 = lender1;
@@ -329,11 +333,9 @@ contract Borrower is IUniswapV3MintCallback {
     }
 
     function getPrices() public view returns (Prices memory prices) {
-        (int24 arithmeticMeanTick, ) = Oracle.consult(UNISWAP_POOL, 1200);
-        uint256 sigma = 0.025e18; // TODO: fetch real data from the volatility oracle
+        (uint160 sqrtMeanPriceX96, uint256 sigma) = ORACLE.consult(UNISWAP_POOL);
 
         // compute prices at which solvency will be checked
-        uint160 sqrtMeanPriceX96 = TickMath.getSqrtRatioAtTick(arithmeticMeanTick);
         (uint160 a, uint160 b) = BalanceSheet.computeProbePrices(sqrtMeanPriceX96, sigma, B);
         prices = Prices(a, b, sqrtMeanPriceX96);
     }
--- a/core/src/VolatilityOracle.sol
+++ b/core/src/VolatilityOracle.sol
@@ -1,0 +1,181 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.17;
+
+import {IUniswapV3Pool} from "v3-core/contracts/interfaces/IUniswapV3Pool.sol";
+
+import {MAX_SIGMA, IV_SCALE, IV_CHANGE_PER_SECOND, ORACLE_LOOKBACK, FEE_GROWTH_GLOBALS_SAMPLE_PERIOD} from "./libraries/constants/Constants.sol";
+import {Oracle} from "./libraries/Oracle.sol";
+import {Volatility} from "./libraries/Volatility.sol";
+
+/// @title VolatilityOracle
+/// @author Aloe Labs, Inc.
+/// @dev "Test everything; hold fast what is good." - 1 Thessalonians 5:21
+contract VolatilityOracle {
+    struct LastWrite {
+        uint8 index;
+        uint32 time;
+        uint216 iv;
+    }
+
+    mapping(IUniswapV3Pool => Volatility.PoolMetadata) public cachedMetadata;
+
+    mapping(IUniswapV3Pool => Volatility.FeeGrowthGlobals[60]) public feeGrowthGlobals;
+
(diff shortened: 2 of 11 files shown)


core/src/Borrower.sol
```diff
@@ -79,7 +82,8 @@ contract Borrower is IUniswapV3MintCallback {
                        CONSTRUCTOR & INITIALIZER
     //////////////////////////////////////////////////////////////*/
 
-    constructor(IUniswapV3Pool pool, Lender lender0, Lender lender1) {
+    constructor(VolatilityOracle oracle, IUniswapV3Pool pool, Lender lender0, Lender lender1) {
```
aneesh1, line 85: Why are we passing the `VolatilityOracle` in as a constructor argument? Previously, we were able to get away without passing the Oracle in so I'm wondering what changed. Also, after #62 I want to be sure this is something we absolutely **need** to add since it'll affect the addresses where we deploy the contracts.

core/src/VolatilityOracle.sol
```diff
@@ -0,0 +1,181 @@
...
+
+/// @title VolatilityOracle
+/// @author Aloe Labs, Inc.
+/// @dev "Test everything; hold fast what is good." - 1 Thessalonians 5:21
+contract VolatilityOracle {
+    struct LastWrite {
+        uint8 index;
+        uint32 time;
+        uint216 iv;
+    }
```
aneesh1, line 18: Nicely packed struct

core/src/VolatilityOracle.sol
```diff
@@ -0,0 +1,181 @@
...
+        // If block times are inconsistent, `maxSecondsAgo` from oracle may be inflated. Divide by 2 to be extra safe.
+        metadata.maxSecondsAgo = Oracle.getMaxSecondsAgo(pool, observationIndex, observationCardinality) / 2;
+        require(metadata.maxSecondsAgo > ORACLE_LOOKBACK, "Aloe: cardinality");
+
+        uint24 fee = pool.fee();
+        metadata.gamma0 = fee;
+        metadata.gamma1 = fee;
+        unchecked {
+            if (feeProtocol % 16 != 0) metadata.gamma0 -= fee / (feeProtocol % 16);
+            if (feeProtocol >> 4 != 0) metadata.gamma1 -= fee / (feeProtocol >> 4);
```
aneesh1, line 99: I understood most of this function. First, we check that the data for the provided `metadata` is older than some time. If it is, we get the instantaneous fee of the `pool` and check if the `feeProtocol` is divisible by 16 and if the `feeProtocol` multiplied by 16 is equal to 0. Why 16?

core/src/VolatilityOracle.sol
```diff
@@ -0,0 +1,181 @@
...
+    mapping(IUniswapV3Pool => Volatility.FeeGrowthGlobals[60]) public feeGrowthGlobals;
+
+    mapping(IUniswapV3Pool => LastWrite) public lastWrites;
+
+    function prepare(IUniswapV3Pool pool) external {
+        cachedMetadata[pool] = _getPoolMetadata(pool);
+
+        if (lastWrites[pool].time == 0) {
+            feeGrowthGlobals[pool][0] = _getFeeGrowthGlobalsNow(pool);
+            lastWrites[pool] = LastWrite({index: 0, time: uint32(block.timestamp), iv: uint216(MAX_SIGMA)});
```
aneesh1, line 31: Initially, we assume the IV is `MAX_SIGMA` - why? How did we determine this value again?

core/src/VolatilityOracle.sol
```diff
@@ -0,0 +1,181 @@
...
+            // Default to using the existing IV
+            uint256 iv = lastWrite.iv;
+            // Only update IV if the feeGrowthGlobals samples are approximately 1 hour apart
+            if (
+                isInInterval({
+                    min: 1 hours - 5 minutes, // NOTE: Keeping these constants here because they're related to `arr.length`
+                    x: b.timestamp - a.timestamp,
+                    max: 1 hours + 5 minutes
+                })
+            ) {
```
aneesh1, line 65: So we're updating IV if the `feeGrowthGlobals` samples are 1 hour apart (technically, we'll do an update at some time between 55 mins to 1 hour 5 mins). I also see that the length of our array is 60 (60 mins in 1 hour) which is why we're keeping this number. Why 1 hour? Even bigger question - are we guaranteed to call update periodically so that we don't miss this window? What happens if we miss this window?

core/src/VolatilityOracle.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    function _getFeeGrowthGlobalsOld(
+        Volatility.FeeGrowthGlobals[60] storage arr,
+        uint256 index
+    ) private view returns (Volatility.FeeGrowthGlobals memory) {
+        uint256 target = block.timestamp - 1 hours;
+
+        // See if the newest sample is best
+        Volatility.FeeGrowthGlobals memory sample = arr[index];
+        if (sample.timestamp <= target) return sample;
```
aneesh1, line 135: `Volatility.FeeGrowthGlobals[60] storage arr` will have multiple values. If the timestamp of the newest sample is less than the `target` time, we return the `sample` because it's best. "Best" in this means "most recent".

core/test/libraries/Volatility.t.sol
```diff
@@ -279,7 +279,7 @@ contract VolatilityTest is Test {
         uint128 tickLiquidity,
         uint32 scale
     ) public view {
-        scale = uint32(bound(scale, 1 minutes, 1 days));
+        scale = uint32(bound(scale, 1 minutes, 2 days));
```
aneesh1, line 282: Why change from 1 days to 2 days?

aneesh1 (commented):
Looks good! Couple of questions - also wondering what the gas savings will be on this (if any).

core/src/Borrower.sol
haydenshively: The alternative is to have the Factory itself create a VolatilityOracle (`new VolatilityOracle()`) in it's constructor, and store it for future use. The problem with that is that it basically appends all of the VolatilityOracle's bytecode to the Factory, and we would exceed the 24kb contract size limit. Conveniently, the VolatilityOracle doesn't depend on anything (no constructor args), so we can deploy it separately and just pass its address into the Factory here.

core/src/VolatilityOracle.sol
haydenshively: Uniswap governance can opt to excise different fees on token0 and token1. The lower 4 bits of `feeProtocol` contain the fee for token0, and the upper 4 bits contain the fee for token1.

core/src/VolatilityOracle.sol
haydenshively: To be extra safe, `Borrower` clamps values received from VolatilityOracle to be between `MIN_SIGMA` and `MAX_SIGMA`. When the VolatilityOracle is first initialized, it's better to be on the upper end of that (more conservative) than the lower end (more risky). But we don't want to set it at some absurdly high value like 2e18 because of the constraint on the derivative (`IV_CHANGE_PER_SECOND`)

core/src/VolatilityOracle.sol
haydenshively: I plan to write a bot similar to the liquidation bot that will call this every hour, on the hour for popular pools. If we miss a window, it's not the end of the world though -- the IVs (and "equivalent collateral factors") would just freeze in place until we start calling it again.

Also, there's an incentive for liquidators to call it when IV goes up (since that brings people closer to liquidation) and there's an incentive for borrowers to call it when IV goes down (since that makes them healthier and/or lets them borrow more). So long-term we shouldn't have to be the only ones calling `update`

core/src/VolatilityOracle.sol
haydenshively: "Best" is intended to mean "nearest to `target`". I'll update the comments.

core/test/libraries/Volatility.t.sol
haydenshively: Referencing our offline discussion: because I changed "N" from 1 hour to 1 day, I wanted to make sure that >1 days was safe in the tests. Thus 2.

core/src/VolatilityOracle.sol
aneesh1: Creating a bot for popular pools sounds good. Also, good to know that we have safety measures in case we miss the window and that calling this is aligned with the liquidator's incentives.

core/src/Borrower.sol
aneesh1: > The problem with that is that it basically appends all of the VolatilityOracle's bytecode to the Factory, and we would exceed the 24kb contract size limit.

TIL! I also think that doing mock testing should be easier with this setup. 👍

haydenshively: Oo good point about mock testing. That's true

PR #64, opened 2023-01-31
Add more fuzz tests for Volatility lib and make revert less likely

haydenshively:


core/src/libraries/Oracle.sol
```diff
@@ -28,16 +28,18 @@ library Oracle {
             secondsAgos
         );
 
-        int56 tickCumulativesDelta = tickCumulatives[1] - tickCumulatives[0];
-        arithmeticMeanTick = int24(tickCumulativesDelta / int32(secondsAgo));
-        // Always round to negative infinity
-        if (tickCumulativesDelta < 0 && (tickCumulativesDelta % int32(secondsAgo) != 0)) arithmeticMeanTick--;
+        unchecked {
```
aneesh1, line 31: Good call moving this to an unchecked block 👍 The deleted lines appear in this block so it looks correct.

core/src/libraries/Volatility.sol
```diff
@@ -80,21 +81,17 @@ library Volatility {
...
             // *at that swap*. But for prices simulated with GBM and swap sizes either normally or uniformly distributed,
             // the error you get from using geometric mean price is <1% even with high drift and volatility.
-            volumeGamma0Gamma1 = revenue1Gamma0 + amount0ToAmount1(revenue0Gamma1, data.arithmeticMeanTick);
-        }
+            uint256 volumeGamma0Gamma1 = revenue1Gamma0 + amount0ToAmount1(revenue0Gamma1, data.arithmeticMeanTick);
 
-        uint128 sqrtTickTVLX32 = uint128(
-            FixedPointMathLib.sqrt(
+            // Fits in uint128
+            uint256 sqrtTickTVLX32 = FixedPointMathLib.sqrt(
```
aneesh1, line 87: If it fits in `uint128`, why do `uint256`? Maybe we either remove the comment or add more detail to the comment?

core/src/libraries/Volatility.sol
```diff
@@ -80,21 +81,17 @@ library Volatility {
...
+            );
+            if (sqrtTickTVLX32 == 0) return 0;
 
-        if (sqrtTickTVLX32 == 0) return 0;
-        unchecked {
-            return
-                (uint256(2e18) * uint256(timeAdjustmentX32) * FixedPointMathLib.sqrt(volumeGamma0Gamma1)) /
-                sqrtTickTVLX32;
+            // Fits in uint48
+            uint256 timeAdjustmentX32 = FixedPointMathLib.sqrt((scale << 64) / (b.timestamp - a.timestamp));
```
aneesh1, line 93: Same as above

core/test/libraries/Volatility.t.sol
```diff
@@ -79,40 +96,13 @@ contract VolatilityTest is Test {
...
                 527278396421895291380335427321388844898052,
-                uint32(block.timestamp)
+                uint32(1660599905)
             ),
             1 days
         );
         assertEq(dailyIV, 0); // 0%
     }
 
-    function testFail_estimate() public pure {
```
aneesh1, line 89: This has been moved to `test_spec_estimate`. Why did this test fail previously? Was it due to revert errors caused by arithmetic underflow/overflow?

core/test/libraries/Volatility.t.sol
```diff
@@ -79,40 +96,13 @@ contract VolatilityTest is Test {
             Volatility.FeeGrowthGlobals(
                 1501955347902231987349614320458936,
                 527278396421895291380335427321388844898052,
-                uint32(block.timestamp)
+                uint32(1660599905)
```
aneesh1, line 99: Why? What's the significance of this timestamp?

core/test/libraries/Volatility.t.sol
```diff
@@ -38,6 +38,23 @@ contract VolatilityTest is Test {
...
+            ),
+            Volatility.FeeGrowthGlobals(
+                1501968291161650295867029090958139,
+                527315901327546020416261134123578344760082,
+                8640
+            ),
+            1 hours
+        );
+        assertEq(dailyIV, 4165347829168579); // 0.417%
+
```
aneesh1, line 57: Good - example of volatility estimate that reverted in the old code, but no longer reverts now

aneesh1 (approved):
Things I checked for
- [x] `if (data.secondsPerLiquidityX128 == 0 || b.timestamp - a.timestamp == 0) return 0;` (line 86 for first part of || line 103 for second)
- [x] Volatility estimate that reverted in the old code, but no longer reverts now (`test_noRevert_computeTickTVLX64`,  `test_noRevert_computeRevenueGamma`, `test_noRevert_estimate`, line 56 in `test_spec_estimate`)

Unsure:
`if (sqrtTickTVLX32 == 0) return 0;`
Couldn't find a test for `if (sqrtTickTVLX32 == 0) return 0;` - I feel like there might be one, I'm just missing it.

core/src/libraries/Volatility.sol
haydenshively: I've seen auditors complain when using memory/stack data types other than `uint256`, because under the hood, all memory numbers are actually `uint256`. So even though it _would_ fit in `uint128`, the (imo silly) convention is to declare it as a `uint256`

core/test/libraries/Volatility.t.sol
haydenshively: Yes

haydenshively:
> Unsure:
> if (sqrtTickTVLX32 == 0) return 0;
> Couldn't find a test for if (sqrtTickTVLX32 == 0) return 0; - I feel like there might be one, I'm just missing it.

Yeah idk if we test this case specifically, but it happens when `tickLiquidity` is so small that the sqrt is 0. We have to return early, otherwise the next line would be division by 0. I did try to take this out, and it broke, so I put it back in

core/test/libraries/Volatility.t.sol
haydenshively: Just a random one :)

PR #65, opened 2023-01-31
Adjust Volatility.PoolData to take a sqrtPrice instead of a tick

haydenshively:


aneesh1 (approved):
Checked function signatures to confirm that instead of a passing a tick, now we're doing `TickMath.getSqrtRatioAtTick(<tick>)`

PR #72, opened 2023-02-17
Add SimpleManager and WithdrawManager

haydenshively:
`SimpleManager` allows for efficient borrows and repays by calling the `Borrower` with `data`. `data` should be an abi-encoded function call. `WithdrawManager` allows for efficient withdrawal from a `Borrower`. `data` should be abi-encoded parameters (amount0, amount1, recipient, calldataForBorrower). If calldataForBorrower is non-empty bytes, `WithdrawManager` will first do a call to `Borrower` (like `SimpleManager`) and _then_ do the withdrawal.


core/.gas-snapshot
```diff
@@ -1,15 +1,15 @@
...
+BorrowerGasTest:test_repay() (gas: 97332)
+BorrowerGasTest:test_uniswapDepositInBorrower() (gas: 293542)
 BorrowerGasTest:test_uniswapDepositStandard() (gas: 167271)
-BorrowerGasTest:test_uniswapWithdraw() (gas: 135376)
-BorrowerGasTest:test_withdraw() (gas: 122372)
+BorrowerGasTest:test_uniswapWithdraw() (gas: 135412)
+BorrowerGasTest:test_withdraw() (gas: 122417)
 FactoryGasTest:test_createBorrower() (gas: 155929)
-FactoryGasTest:test_createMarket() (gas: 3657871)
+FactoryGasTest:test_createMarket() (gas: 3450501)
```
haydenshively, line 12: No core code changed, and we're still using Solidity 0.8.17, so afaict these changes in gas are due to upgrading foundry.

core/foundry.toml
```diff
@@ -13,6 +13,9 @@ optimizer_runs = 1000000
 gas_reports = []
 
 auto_detect_remappings = true
+remappings = [
+    'openzeppelin-contracts/=lib/openzeppelin-contracts/contracts/',
+]
```
haydenshively, line 18: Had to add this as a result of foundry upgrade as well

IanWoodard (approved):
LGTM

PR #74, opened 2023-02-21
Create UniswapNFTManager so users can import NFTs as collateral

haydenshively:


periphery/src/managers/UniswapNFTManager.sol
```diff
@@ -0,0 +1,92 @@
...
+        );
+
+        NFT_MANAGER.collect(
+            INFTManager.CollectParams({
+                tokenId: tokenId,
+                recipient: recipient,
+                amount0Max: type(uint128).max,
+                amount1Max: type(uint128).max
+            })
+        );
```
IanWoodard, line 90: I assume we opt not to burn it to increase liquidity later on. Does this sound right?

periphery/src/managers/UniswapNFTManager.sol
```diff
@@ -0,0 +1,92 @@
...
+            NFT_MANAGER.increaseLiquidity(
+                INFTManager.IncreaseLiquidityParams({
+                    tokenId: tokenId,
+                    amount0Desired: burned0,
+                    amount1Desired: burned1,
+                    amount0Min: 0,
+                    amount1Min: 0,
+                    deadline: block.timestamp
+                })
+            );
```
IanWoodard, line 68: The reason we don't burn the NFT, correct?

periphery/src/managers/UniswapNFTManager.sol
```diff
@@ -0,0 +1,92 @@
...
+        if (liquidity < 0) {
+            // safety checks since this contract will be approved to manager users' positions
+            {
+                require(FACTORY.isBorrower(msg.sender));
+                (address owner, , ) = borrower.slot0();
+                require(owner == NFT_MANAGER.ownerOf(tokenId));
+            }
+
+            _withdrawFromNFT(tokenId, uint128(-liquidity), msg.sender);
+            borrower.uniswapDeposit(lower, upper, uint128(-liquidity));
```
IanWoodard, line 46: Should this be negative?

IanWoodard (approved):

periphery/src/managers/UniswapNFTManager.sol
haydenshively: Exactly!

periphery/src/managers/UniswapNFTManager.sol
haydenshively: Yes. The sign of `liquidity` determines whether we move liquidity from NFT to Borrower, or Borrower to NFT. Negative means into Borrower, but once we've checked the sign all the actual deposit logic expects a `uint128`.

PR #80, opened 2023-04-05
Give Factory control of ante and nSigma

haydenshively:
In contrast to #78, this allows the `Factory` to update parameters for all Borrowers in a given pool simultaneously. For example, if it is determined that pool ABC/XYZ is systematically under-reporting volatility for some reason, the `Factory` could increase `nSigma` for that pool to be safer.

#78 would've required parameters to be changed individually for each Borrower.

Note that this results in ~1000 gas increase for all Borrower methods, but that's the price we pay for upgradeability.


core/src/Factory.sol
```diff
@@ -27,6 +29,11 @@ contract Factory {
         Borrower borrowerImplementation;
     }
 
+    struct Parameters {
+        uint248 ante;
+        uint8 nSigma;
```
aneesh1, line 34: The `ANTE` constant used to be a `uint256`, now it's a `uint248`. Is it OK to lose the 8-bits?

core/src/Factory.sol
```diff
@@ -35,6 +42,8 @@ contract Factory {
 
     mapping(IUniswapV3Pool => Market) public getMarket;
 
+    mapping(IUniswapV3Pool => Parameters) public getParameters;
```
aneesh1, line 45: Are `mapping`s modifiable? Or will we need some kind of setter method in a future PR to modify the `Parameters`?

aneesh1 (approved):
I understand that in your previous approach, the parameters were configured per `Borrower`. It sounds like we would have had to force consistency of all borrowers across the pool to make sure they all had the same `Parameters`, so this is definitely a better approach.

It makes sense that all `Borrower`s in a particular pool should have the same `ANTE` and `nSigma`. Especially given that we're using `nSigma` to `computeProbePrices`, different pools should have different probe prices.

core/src/Factory.sol
haydenshively: We will need a setter method in a future PR. I'm preparing things to be governable before actually making them governable, if that makes sense.

core/src/Factory.sol
haydenshively: Yeah I don't expect `ANTE` to be larger than 1 ETH, which is ceil(log2(1e18)) = 60 bits, so 248 is more than enough.

PR #83, opened 2023-04-11
Bring back flash loan functionality that was removed before beta launch

haydenshively:
Flash loan functionality was removed [here](https://github.com/aloelabs/aloe-ii/commit/50959f12bf4b9b1973b86bbccf62f6d59fd8dd74#diff-bd82376927d64137b16c2eb48179808b927a9644b06517dfb511e02bdce676ec) because it hadn't been fully tested.

Now I've brought it back with a few sanity checks:
- Implemented from scratch and compared with what we had before -- got basically the same code
- Made sure that (if we wanted to) we could create an adapter to conform to [ERC3156](https://eips.ethereum.org/EIPS/eip-3156). See below for what this might look like:

```Solidity
contract ERC3156Adapter is IFlashBorrower {
  Lender public immutable LENDER;

  ERC20 public immutable ASSET;

  constructor(Lender lender) {
    LENDER = lender;
    ASSET = lender.asset();
  }

  function maxFlashLoan(ERC20 token) external view returns (uint256) {
    if (token == ASSET) return token.balanceOf(address(LENDER));
    return 0;
  }

  function flashFee(address token, uint256) external view returns (uint256) {
    require(token == ASSET);
    return 0;
  }

  function flashLoan(
    IERC3156FlashBorrower receiver,
    address,
    uint256 amount,
    bytes calldata data
  ) external returns (bool) {
    LENDER.flash(amount, this, abi.encode(receiver, data));
    return true;
  }

  function onFlashLoan(address initiator, uint256 amount, bytes calldata allData) external {
    (IERC3156FlashBorrower receiver, bytes memory data) = abi.decode(allData, (IERC3156FlashBorrower, bytes));

    ASSET.safeTransfer(address(receiver, amount));

    require(
      receiver.onFlashLoan(initiator, ASSET, amount, 0, data) == keccak256("ERC3156FlashBorrower.onFlashLoan"),
      "IERC3156: Callback failed"
    );

    ASSET.safeTransferFrom(address(receiver), address(LENDER), amount);
  }
}
```

I have explicitly chosen _not_ to include this adapter in the repo because I do not think it's the best way to receive a flash loan from Aloe II. The extra functions like `maxFlashLoan` and `flashFee` are no big deal, but unfortunately ERC3156 requires the lender to use a pull architecture. As noted in [the final paragraph of the "Rationale" section](https://eips.ethereum.org/EIPS/eip-3156#rationale), this is convenient for most protocols. But for Aloe is one of the few that's based entirely on a push architecture, so for us it's just a waste of gas.


core/src/Lender.sol
```diff
@@ -210,6 +210,23 @@ contract Lender is Ledger {
         emit Repay(msg.sender, beneficiary, amount, units);
     }
 
+    /// @dev Reentrancy guard is critical here! Without it, one could use a flash loan to repay a normal loan.
+    function flash(uint256 amount, IFlashBorrower to, bytes calldata data) external {
+        // Guard against reentrancy
+        uint32 lastAccrualTime_ = lastAccrualTime;
+        require(lastAccrualTime_ != 0, "Aloe: locked");
+        lastAccrualTime = 0;
```
aneesh1, line 218: This looks good. Ran through this chain of events in my head:
1. Call `flash`
2. `lastAccuralTime_` is set
3. Passes the require statement
4. `lastAccuralTime` is set to 0
5. `onFlashLoan` in the `IFlashBorrower` is triggered
6. Within `onFlashLoan`, we try to call `flash` again 
7. When we re-enter `flash`, `lastAccrualTime_ = 0` 
8. We fail the `require` check

Seems to me like this is will defend us from a re-entrancy attack.

aneesh1 (approved):
Looks good!

core/src/Lender.sol
haydenshively: Yep! Note on (6), one of the other functions could be called too -- like `deposit` or `repay`. We defend against reentrancy there in the same way: the `_load` function checks that `lastAccrualTime != 0`

PR #84, opened 2023-04-11
Tweak volatility oracle constants for more robust feeGrowthGlobal sampling

haydenshively:
Builds on #83 so let's get that merged first.


core/src/VolatilityOracle.sol
```diff
@@ -60,9 +60,9 @@ contract VolatilityOracle {
             // Only update IV if the feeGrowthGlobals samples are approximately 1 hour apart
             if (
                 isInInterval({
-                    min: 1 hours - 5 minutes, // NOTE: Keeping constants in-line because they're related to arr.length
+                    min: FEE_GROWTH_AVG_WINDOW - 3 * FEE_GROWTH_SAMPLE_PERIOD,
```
aneesh1, line 63: Why `- 3 * FEE_GROWTH_SAMPLE_PERIOD`?

core/src/VolatilityOracle.sol
```diff
@@ -60,9 +60,9 @@ contract VolatilityOracle {
             // Only update IV if the feeGrowthGlobals samples are approximately 1 hour apart
```
aneesh1: Adjust comment?

``Only update IV if the feeGrowthGlobals samples are approximately FEE_GROWTH_AVG_WINDOW hour apart``

core/src/libraries/constants/Constants.sol
```diff
@@ -29,6 +29,20 @@ uint256 constant IV_SCALE = 24 hours;
...
+uint256 constant FEE_GROWTH_ARRAY_LENGTH = 72;
+
+/// @dev The minimum number of seconds that must elapse before a new feeGrowthGlobals sample will be stored. This also
+/// controls how often the oracle can update IV.
+uint256 constant FEE_GROWTH_SAMPLE_PERIOD = 5 minutes;
+
+/// @dev To compute Uniswap mean price & liquidity, we need 2 samples. One is always at the current block, the other is
+/// from `UNISWAP_AVG_WINDOW` seconds ago. Larger values make the resulting price/liquidity values harder to
+/// manipulate, but also make the oracle slower to respond to changes.
+uint32 constant UNISWAP_AVG_WINDOW = 30 minutes;
```
aneesh1, line 48: I'm curious, how did you come up with these? I'm assuming the effect increasing the `FEE_GROWTH_SAMPLE_PERIOD` from 1 minute to 5 minutes as well as the `UNISWAP_AVG_WINDOW` from 20 minutes to 30 minutes make it more robust against spikes in price, but I'm wondering why these values in particular?

aneesh1 (approved):
Looks good! I trust that these make the `feeGrowthGlobal` sampling more robust, just curious why you chose 5 minutes for the `FEE_GROWTH_SAMPLE_PERIOD` and 30 minutes for the `UNISWAP_AVG_WINDOW`.

core/src/VolatilityOracle.sol
haydenshively: Well ideally the samples would be placed `FEE_GROWTH_AVG_WINDOW` hours apart, but I wanted to provide some wiggle room. +/- 15 minutes on a 6 hour window felt about right, and is certainly achievable based on the consistency of block timestamps.

core/src/libraries/constants/Constants.sol
haydenshively: I changed `UNISWAP_AVG_WINDOW` to 30 minutes based on Euler's research. I increased `FEE_GROWTH_SAMPLE_PERIOD` to 5 minutes so that we could keep the ring buffer a reasonable length while extending the `FEE_GROWTH_AVG_WINDOW` to 6 hours. I wanted 6 hours on that one to reduce the likelihood of sampling over an hour in which there was no trading activity.

PR #86, opened 2023-04-24
Fully support Permit2 in Router and add function that can handle dynamic maxRedeem

haydenshively:


periphery/src/Router.sol
```diff
@@ -105,29 +67,55 @@ contract Router {
...
 
-    function repayWithApprove(Lender lender, uint256 amount, address beneficiary) external returns (uint256 units) {
-        lender.asset().safeTransferFrom(msg.sender, address(lender), amount);
+    function repayWithPermit2(
+        Lender lender,
+        uint256 amount,
+        address beneficiary,
+        uint256 nonce,
+        uint256 deadline,
+        bytes calldata signature
```
aneesh1, line 76: I might not have understood the `Permit2` docs correctly, so please correct me if I'm wrong. Clients of this function must call `ERC20(token).approve(Permit2)`. Is this `Permit2` Contract the same one that has been passed to the `Router`?

periphery/src/Router.sol
```diff
@@ -105,29 +67,55 @@ contract Router {
...
+            // The packed signature that was the result of signing
+            // the EIP712 hash of `permit`.
+            signature
+        );
+
         units = lender.repay(amount, beneficiary);
     }
 
-    function repayWithPermit(
+    function isMaxRedeemDynamic(Lender lender, address owner) external view returns (bool) {
```
aneesh1, line 100: When might this function be used?

periphery/src/Router.sol
```diff
@@ -105,29 +67,55 @@ contract Router {
...
+        uint256 maxRedeem = lender.maxRedeem(msg.sender);
+        if (shares > maxRedeem) shares = maxRedeem;
 
-        if (allowance != 0) {
-            asset.permit(msg.sender, address(this), allowance, deadline, v, r, s);
-        }
-
-        asset.safeTransferFrom(msg.sender, address(lender), amount);
-        units = lender.repay(amount, beneficiary);
+        lender.permit(msg.sender, address(this), shares, deadline, v, r, s);
```
aneesh1: This looks like you're using `Permit` from [EIP-2612](https://github.com/dragonfly-xyz/useful-solidity-patterns/tree/main/patterns/permit2#permit-eip-2612-model) instead of `Permit2`. Is that correct? If so, the link provided mentions that this isn't supported by all major tokens:

> Since EIP-2612 is an extension of the ERC20 standard, this functionality is only possible on new (or upgradeable) tokens. So there are very few major tokens in the wild where this pattern actually works.

Should this also be using `Permit2`?

aneesh1 (commented):
Read over some documentation [here](https://github.com/dragonfly-xyz/useful-solidity-patterns/tree/main/patterns/permit2#permit-eip-2612-model) prior to reviewing. I have some questions about how clients ought to use `repayWithPermit2` and a question about the `redeemWithChecks` function.

periphery/src/Router.sol
haydenshively: Correct.

periphery/src/Router.sol
haydenshively: Used to determine whether a user's `maxRedeem` number of shares is changing. If it is, we may use the helper function on `Router` to redeem instead of plain `lender.redeem(...)`

periphery/src/Router.sol
haydenshively: Yes, this is using EIP2612. The `Router` needs permission to burn the user's shares. Since that permission is governed by our `Lender` contract (which implements ERC20 and the EIP2612 extension) we don't need to worry about compatibility.

aneesh1 (approved):

haydenshively:
@aneesh1 could you re-approve this please? Github seems not to have noticed that you approved it before, and won't let me merge.

aneesh1 (approved):

aneesh1 (approved):

aneesh1:
It's still saying "Merging is blocked", but I have no idea why. As a workaround, maybe we request @IanWoodard's review and then he can approve to unblock?

IanWoodard (approved):
LGTM

PR #87, opened 2023-05-03
Add transferFrom option to deposit to conform with ERC4626

haydenshively:
Part of #33 

This change allows users to deposit directly, without going through any kind of Router (as long as they've `approve`d the `Lender` to spend their tokens). This is required by the [ERC4626 standard](https://eips.ethereum.org/EIPS/eip-4626#methods):

> MUST support EIP-20 approve / transferFrom on asset as a deposit flow. MAY support an additional flow in which the underlying tokens are owned by the Vault contract before the deposit execution, and are accounted for during deposit.

The way I've done it here, users must choose to _either_ prepay _or_ use the `transferFrom` flow. Mixing and matching (partial prepay with remainder being covered by `transferFrom`) won't work. An alternative implementation would allow for mixing and matching, but @IanWoodard and I decided it was less readable:

_(replacing lines 124-128)_
```Solidity
uint256 prepayAmount = asset_.balanceOf(address(this)) - cache.lastBalance;
if (prepayAmount < amount) {
    asset_.safeTransferFrom(msg.sender, address(this), amount - prepayAmount);
}
cache.lastBalance += amount;
```


core/src/libraries/Positions.sol
```diff
@@ -29,7 +29,6 @@ function zip(int24[6] memory positions) pure returns (uint144 zipped) {
  * positions like [-100, 100, -100, 100].
  * @param zipped Encoded Uniswap positions. Equivalent to the layout of `int24[6] storage yourPositions`
  * @return positionsOfNonZeroWidth Flattened array of Uniswap positions that may or may not hold liquidity
- * TODO: consider simply checking whether ticks are strictly increasing. if they are, they must be unique and we can save gas elsewhere
```
haydenshively, line 32: Unrelated to the deposit stuff, but I considered this and realized it was dumb. If ticks were strictly increasing, it'd be impossible to have a narrow position (like [100, 200]) inside of a wide position (like [-5678, 1234]). So the current logic is good.

core/src/Lender.sol
```diff
@@ -119,9 +119,13 @@ contract Lender is Ledger {
         shares = _convertToShares(amount, inventory, cache.totalSupply, /* roundUp: */ false);
         require(shares != 0, "Aloe: zero impact");
 
-        // Ensure tokens were transferred
+        // Ensure tokens are transferred
+        ERC20 asset_ = asset();
         cache.lastBalance += amount;
-        require(cache.lastBalance <= asset().balanceOf(address(this)), "Aloe: insufficient pre-pay");
+        bool didPrepay = cache.lastBalance <= asset_.balanceOf(address(this));
```
aneesh1, line 125: I understand now - if the user pre-pays using the `Router`, then that means that [this check](https://github.com/aloelabs/aloe-ii/pull/87/files#diff-bd82376927d64137b16c2eb48179808b927a9644b06517dfb511e02bdce676ecR125) will return `true`, thus only pre-pay is supported through this flow. ✅

core/src/Lender.sol
```diff
@@ -119,9 +119,13 @@ contract Lender is Ledger {
...
+        if (!didPrepay) {
+            asset_.safeTransferFrom(msg.sender, address(this), amount);
+        }
```
aneesh1, line 128: User did not pre-pay using the `Router`, so we call `transferFrom` on the `asset` which satisfies this requirement: 
> MUST support EIP-20 approve / transferFrom on asset as a deposit flow. 

✅

aneesh1 (approved):
Looks good! I think this is doing what you want it to do and conforming to the standards. I tried to think of a use-case for the "mixing and matching" that you described, but couldn't think of anything.

IanWoodard (approved):
I still think we made the right call not allowing for the mixing and matching of payments as it just adds extra complexity without any obvious use case. LGTM

PR #88, opened 2023-05-03
Add mint and withdraw functions for ERC4626

haydenshively:
These functions are required by ERC4626. Instead of creating additional bespoke logic, they just call the existing functions (`mint` --> `deposit` and `withdraw` --> `redeem`). As such, they will require substantially more gas and **are not recommended** for frontend usage. Literally only here for potential ERC4626 integrators.


aneesh1 (approved):

core/src/Lender.sol
```diff
@@ -137,7 +137,13 @@ contract Lender is Ledger {
         emit Deposit(msg.sender, beneficiary, amount, shares);
     }
 
-    function redeem(uint256 shares, address recipient, address owner) external returns (uint256 amount) {
+    // TODO: We aren't `require`ing that the return value of `deposit` == `shares`. Check it in fuzz tests.
```
IanWoodard, line 140: Could you expand on this?

IanWoodard (approved):

core/src/Lender.sol
haydenshively: In theory, we could add an additional `require` that the return value of deposit is equal to the original passed-in `shares`. But as long as the `previewMint` function is up to spec, that's guaranteed anyway. So instead of adding the `require` and wasting gas, I added the TODO for us to verify `previewMint` functionality with fuzz testing. (We actually already test `previewDeposit` and `previewRedeem` so it won't be too much work to add `previewMint` and `previewWithdraw`)

PR #90, opened 2023-05-09
Add ERC4626 fuzz tests

haydenshively:
Assertions are based on the [official spec](https://eips.ethereum.org/EIPS/eip-4626)

Biggest downside of having these invariants separate from the `Lender` invariants is that they don't get tested with borrows present. I left some TODOs for myself to combine `ERC4626Harness` and `LenderHarness`, so that everything gets tested together.

### DIFF
--- a/core/test/invariants/ERC4626Invariants.t.sol
+++ b/core/test/invariants/ERC4626Invariants.t.sol
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.17;
+
+import "forge-std/Test.sol";
+import "forge-std/Vm.sol";
+
+import {ClonesWithImmutableArgs} from "clones-with-immutable-args/ClonesWithImmutableArgs.sol";
+import {ERC4626, ERC20} from "solmate/mixins/ERC4626.sol";
+import {MockERC20} from "solmate/test/utils/mocks/MockERC20.sol";
+
+import "src/Lender.sol";
+
+import {ERC4626Harness} from "./ERC4626Harness.sol";
+
+contract ERC4626InvariantsTest is Test {
+    ERC20 public asset;
+
+    ERC4626 public vault;
+
+    ERC4626Harness public vaultHarness;
+
+    struct ThingsThatShouldntChange {
+        string name;
+        string symbol;
+        uint8 decimals;
+        ERC20 asset;
+        bytes32 domainSeparator;
+    }
+
+    ThingsThatShouldntChange public thingsThatShouldntChange;
+
+    function setUp() public {
+        {
+            asset = new MockERC20("Token", "TKN", 18); // TODO: replace 18 with an env var
+            address lenderImplementation = address(new Lender(address(2)));
+            Lender lender = Lender(ClonesWithImmutableArgs.clone(
+                lenderImplementation,
+                abi.encodePacked(address(asset))
+            ));
+            RateModel rateModel = new RateModel();
+            lender.initialize(rateModel, 8); // TODO: replace 8 with an env var
+
+            vault = ERC4626(address(lender));
+            vaultHarness = new ERC4626Harness(lender);
+
+            targetContract(address(vaultHarness));
+
+            // forge can't simulate transactions from addresses with code, so we must exclude all contracts
+            excludeSender(address(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D)); // vm
+            excludeSender(address(0x4e59b44847b379578588920cA78FbF26c0B4956C)); // built-in create2 deployer
+            excludeSender(address(this));
+            excludeSender(address(asset));
+            excludeSender(address(lenderImplementation));
+            excludeSender(address(rateModel));
+            excludeSender(address(vault));
+            excludeSender(address(vaultHarness));
+        }
+
+        thingsThatShouldntChange = ThingsThatShouldntChange(
+            vault.name(),
+            vault.symbol(),
+            vault.decimals(),
+            vault.asset(),
+            vault.DOMAIN_SEPARATOR()
+        );
+    }
+
+    function invariant_thingsThatShouldntChangeDontChange() public {
+        ThingsThatShouldntChange memory update = ThingsThatShouldntChange(
+            vault.name(),
+            vault.symbol(),
+            vault.decimals(),
+            vault.asset(),
+            vault.DOMAIN_SEPARATOR()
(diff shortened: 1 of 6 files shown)


core/test/invariants/LenderHarness.sol
```diff
@@ -161,7 +166,7 @@ contract LenderHarness {
         // Make sure `msg.sender` has enough assets to deposit
         if (amountToTransfer > 0) {
             vm.prank(msg.sender);
-            vm.expectRevert(bytes(shares > 0 ? "Aloe: insufficient pre-pay" : "Aloe: zero impact"));
+            vm.expectRevert(bytes(shares > 0 ? "TRANSFER_FROM_FAILED" : "Aloe: zero impact"));
```
haydenshively, line 169: Forgot to update this when we added the `approve` flow in https://github.com/aloelabs/aloe-ii/pull/87

core/test/invariants/ERC4626Invariants.t.sol
```diff
@@ -0,0 +1,110 @@
...
+        );
+
+        assertEq(bytes(update.name), bytes(thingsThatShouldntChange.name));
+        assertEq(bytes(update.symbol), bytes(thingsThatShouldntChange.symbol));
+        assertEq(update.decimals, thingsThatShouldntChange.decimals);
+        assertEq(address(update.asset), address(thingsThatShouldntChange.asset));
+        assertEq(update.domainSeparator, thingsThatShouldntChange.domainSeparator);
+    }
+
+    function invariant_conversionsAreInverses() public {
```
aneesh1, line 84: In the EIP for ERC-4646, `convertToShares` & `convertToAssets` contain the following specifications:
- [ ] MUST NOT be inclusive of any fees that are charged against assets in the Vault.
- [ ] MUST NOT show any variations depending on the caller
- [ ] MUST NOT reflect slippage or other on-chain conditions, when performing the actual exchange.
- [ ] MUST NOT revert unless due to integer overflow caused by an unreasonably large input.
- [ ] MUST round down towards 0.
- [ ] MAY NOT reflect the “per-user” price-per-share
- [ ] Should reflect the “average-user’s” price-per-share

Do we already test these conditions? If so, where? If not, do we plan to include these in a future PR or are they unnecessary?

core/test/invariants/ERC4626Invariants.t.sol
```diff
@@ -0,0 +1,110 @@
...
+        }
+        assertEq(totalSupply, vault.totalSupply());
+    }
+
+    function invariant_maxRedeemLessThanBalance() public {
+        uint256 count = vaultHarness.getHolderCount();
+        for (uint256 i = 0; i < count; i++) {
+            address user = vaultHarness.holders(i);
+
+            assertLe(vault.maxRedeem(user), vault.balanceOf(user));
```
aneesh1, line 107: Nice assert  🙌  

I think this demonstrates how we're complying with certain requirements like this one:

> MUST return the maximum amount of shares that could be transferred from owner through redeem and not cause a revert, which MUST NOT be higher than the actual maximum that would be accepted (it should underestimate if necessary).

ERC4646 goes a step further to even include this requirement to indicate what should happen when limits are imposed. 

> MUST factor in both global and user-specific limits, like if redemption is entirely disabled (even temporarily) it MUST return 0.

Can limits be imposed from the contracts in Aloe II such that 0 will be returned? If so, do we need some test to demonstrate this?

aneesh1 (approved):
Looks good! The ERC 4646 doc does mention **a lot** of invariants. Do we plan to add more invariant tests in future PRs?

core/test/invariants/ERC4626Invariants.t.sol
haydenshively: I think these requirements fall into 2 categories: those which (I believe) are impossible to test, and those which are impossible to test in the invariants file, but could be added to the harness. If you can think of a way to test the ones I've labeled "Impossible", please let me know!

## Impossible

> MUST NOT be inclusive of any fees that are charged against assets in the Vault.

I don't know where we'd get the counterfactual against which to test this. In other words, there's no ERC4626 method that's like `convertToShares` but _inclusive_ of fees, so it's impossible to tell whether this version is _not inclusive_ of fees.

> MUST NOT reflect slippage or other on-chain conditions, when performing the actual exchange.

Again, very difficult to come up with the counterfactual using only the exposed ERC4626 view methods. This is another one where I'm satisfied with proof by inspection -- we stare at the implementation for a while until we're convinced that it's up to spec.

> MUST round down towards 0.

Same issue. How would we be able to tell without inspecting the implementation?

## Add to harness

> MUST NOT show any variations depending on the caller (_and_ MAY NOT reflect the “per-user” price-per-share _and_ Should reflect the “average-user’s” price-per-share)

In forge's invariant test setting, the `function invariant_blabla()` methods cannot have input args. So as much as I want to test this by calling the function multiple times with different pranked addresses (`vm.prank(randomAddress)`), I can't. However, I should be able to create a view-only harness function that forge _will_ call with the desired random args.

> MUST NOT revert unless due to integer overflow caused by an unreasonably large input.

I can do the same here.

core/test/invariants/ERC4626Invariants.t.sol
haydenshively: In the harness, we use `vault.maxRedeem(user)` as an upper bound on the random/fuzzed redemption amount. This implies that (sometimes), the redemption amount will exactly equal `vault.maxRedeem(user)`. The harness will then test like so:

- The value from `maxRedeem` corresponds to some amount of underlying asset. If that amount is 0, assert that it reverts with `"Aloe: zero impact"`
- Otherwise, assert successful transaction, appropriate event emission, and correct balance changes

So to answer your question: no, we don't have any kind of emergency/disable functionality right now. But if we did, I think the harness would be sufficient to check that `maxRedeem` is accurate and precise at the user level. And if it's good at the user level _for all users_, that should imply it's good globally.

haydenshively:
> Looks good! The ERC 4646 doc does mention **a lot** of invariants. Do we plan to add more invariant tests in future PRs?

It really does have a lot. I'll admit the invariant tests are a little light -- I don't know how to express many of them without some form of code introspection. That said, the harness does a lot of heavy lifting for us, even though those assertions aren't "invariants" in the foundry sense of the word.

While this doesn't prove we conform to the standard, I'm at least satisfied that it'll catch any regressions as I change things in the coming weeks.

IanWoodard (approved):

PR #91, opened 2023-05-10
Test that Positions library is memory safe

haydenshively:
forge vm added an `expectSafeMemory` [cheat](https://github.com/foundry-rs/foundry/pull/4491). Taking advantage of that for new tests.

### DIFF
--- a/core/test/libraries/Positions.t.sol
+++ b/core/test/libraries/Positions.t.sol
@@ -3,7 +3,7 @@ pragma solidity 0.8.17;
 
 import "forge-std/Test.sol";
 
-import {Positions, zip} from "src/libraries/Positions.sol";
+import {Positions, extract, zip} from "src/libraries/Positions.sol";
 
 contract PositionsTest is Test {
     using Positions for int24[6];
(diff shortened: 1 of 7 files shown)


core/test/libraries/Positions.t.sol
```diff
@@ -19,6 +19,57 @@ contract PositionsTest is Test {
...
+            let ptr := mload(0x40)
+            mstore(ptr, dirt)
+            mstore(add(ptr, 0x20), dirt)
+            mstore(add(ptr, 0x40), dirt)
+            mstore(add(ptr, 0x60), dirt)
+            mstore(add(ptr, 0x80), dirt)
+            // ...could add more, doesn't really matter
+        }
+
+        int24[] memory extracted = extract(zipped);
```
haydenshively, line 59: Goal is to make sure `extract` still works, even when the memory it's writing to is dirty (see this [big warning](https://docs.soliditylang.org/en/v0.8.19/internals/layout_in_memory.html))

core/test/libraries/Positions.t.sol
```diff
@@ -19,6 +19,57 @@ contract PositionsTest is Test {
...
+        uint64 ptr;
+        assembly ("memory-safe") {
+            ptr := mload(0x40)
+        }
+
+        vm.expectSafeMemory(ptr, ptr + 32 * (1 + 2 * count));
+        extract(zipped);
+    }
+
+    function test_memoryExtractDirty(int24 xl, int24 xu, uint256 dirt) public {
```
aneesh1, line 44: Good test 👍

core/test/libraries/Positions.t.sol
```diff
@@ -19,6 +19,57 @@ contract PositionsTest is Test {
...
+        if (yl != yu) count += 1;
+        if (zl != zu) count += 1;
+        uint256 zipped = zip([xl, xu, yl, yu, zl, zu]);
+
+        uint64 ptr;
+        assembly ("memory-safe") {
+            ptr := mload(0x40)
+        }
+
+        vm.expectSafeMemory(ptr, ptr + 32 * (1 + 2 * count));
```
aneesh1, line 40: Took some time to wrap my head around this, but this makes sense. Start at the free memory pointer and then move in 32 byte increments depending on how many elements there are in our compressed data format (which is just one number). 👍

aneesh1 (approved):
Looks good!

IanWoodard (approved):
While this does not prove the Positions library is memory safe, it is a good starting point and gives us more confidence than before.

PR #92, opened 2023-05-16
TickMath memory safety

haydenshively:
Uniswap's TickMath library had a ton of `assembly` blocks for no clear reason. I merged them together, labelled them "memory-safe", and tested the memory safety in fuzz tests.

### DIFF
--- a/core/test/libraries/TickMath.t.sol
+++ b/core/test/libraries/TickMath.t.sol
@@ -8,6 +8,14 @@ import {TickMath} from "src/libraries/TickMath.sol";
 contract TickMathTest is Test {
     function setUp() public {}
 
+    function test_memoryGetTickAtSqrtRatio(uint160 sqrtPriceX96) public {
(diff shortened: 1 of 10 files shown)


core/test/libraries/TickMath.t.sol
```diff
@@ -57,4 +65,167 @@ contract TickMathTest is Test {
...
+
+        int24 a = TickMath.getTickAtSqrtRatio(sqrtPriceX96);
+        int24 b = getTickAtSqrtRatio(sqrtPriceX96);
+
+        assertEq(a, b);
+    }
+}
+
+/// @dev Original Uniswap implementation
+function getTickAtSqrtRatio(uint160 sqrtPriceX96) pure returns (int24 tick) {
```
aneesh1, line 88: Great idea 💯

aneesh1 (approved):
Seems like there's a 1:1 mapping between the lines removed from the disjoint `assembly` blocks and those added to the contiguous `assembly ("memory-safe")` block. Nice 👏

core/test/libraries/TickMath.t.sol
haydenshively: Gracias :)

PR #93, opened 2023-05-17
Simplify Lender._save function

### LINKED ISSUE #42: Revisit `Lender._save` once we have more comprehensive tests
My comment in `_save` implies that we're assigning `lastAccrualTime` to `cache.lastAccrualTime` instead of `block.timestamp` because the one in `cache` may be older. But afaict, the only time that's the case is if `previewInterest` didn't run due to there being no borrows. Seems like it'd be ok to remove that case and run `previewInterest` anyway. If so, then we could simplify the branches in `_save`.

Need to be more certain this won't break anything though, so wait for additional Lender tests.

haydenshively:
Fixes #42


IanWoodard (approved):
LGTM

aneesh1 (approved):

PR #94, opened 2023-06-01
Add ERC20 staking rewards for Lender

haydenshively:
Implements ERC20 staking rewards. Similar to [Yield Protocol](https://github.com/yieldprotocol/yield-utils-v2/blob/main/src/token/ERC20Rewards.sol) except we store a compressed version of `totalSupply` in the PoolState rewards struct. This eliminates unnecessary storage reads/writes, increasing gas efficiency.

Things that are missing:
- [ ] Methods that allow governance to set rewards rate and add/remove `REWARDS_TOKEN` to the pool
- [x] Method(s) that allow users to actually claim the rewards they've earned

I'm going to hold off on implementing the governance stuff for when we actually have governance contracts. The framework is there anyway; it'll just be a matter of exposing `Rewards.setRate` to external callers.


core/src/Lender.sol
```diff
@@ -448,6 +481,6 @@ contract Lender is Ledger {
 
         totalSupply = cache.totalSupply.safeCastTo112();
         lastBalance = cache.lastBalance.safeCastTo112();
-        lastAccrualTime = block.timestamp.safeCastTo32(); // Disables reentrancy guard
+        lastAccrualTime = uint32(block.timestamp); // Disables reentrancy guard
```
haydenshively, line 495: Unrelated tweak

core/src/libraries/Rewards.sol
```diff
@@ -0,0 +1,122 @@
...
+/// @title Rewards
+/// @notice Implements logic for staking rewards
+/// @author Aloe Labs, Inc.
+/// @author Inspired by Yield Protocol (https://github.com/yieldprotocol/yield-utils-v2/blob/main/src/token/ERC20Rewards.sol)
+library Rewards {
+    bytes32 private constant REWARDS_SLOT = keccak256("aloe.ii.rewards");
+
+    struct PoolState {
+        uint168 accumulated; // Accumulated rewards per token for the period, scaled up by 1e18
+        uint32 lastUpdated; // Last time the rewards per token accumulator was updated
```
IanWoodard: This code will only work for the next ~80 years. I understand that this can help us pack things tightly and save gas, but just making a note of it.

core/src/libraries/Rewards.sol
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    /// @dev Adjusts `rate` to account for changes in `totalSupply`
+    function _rate(uint56 rate, uint256 oldTotalSupply, uint256 newTotalSupply) private pure returns (uint56) {
+        unchecked {
+            if (oldTotalSupply == 0) oldTotalSupply = 1;
+            if (newTotalSupply == 0) newTotalSupply = 1;
+
+            return uint56((rate * oldTotalSupply) / newTotalSupply);
```
IanWoodard: Does it make sense to return a non-zero rate if there is no supply?

IanWoodard: Also, wouldn't the rate be disproportionate if the total supply is zero and new tokens are minted?

IanWoodard (commented):
Looks good so far. Left a few comments. Also, I am interested to see how the claiming process will work.

core/src/libraries/Rewards.sol
haydenshively: Yep, we're already using uint32 timestamps in other parts of the protocol, so this isn't a new limitation. Will likely get called out in audits though 🙃

core/test/libraries/Log2.t.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    function test_log2_range(uint256 x) public {
+        vm.assume(x > 0);
+
+        int256 y = log2(x);
+        assertGe(y, -128 << 10);
+        assertLt(y, 128 << 10);
+    }
+
+    function test_log2_match(uint256 x) public {
```
aneesh1, line 51: Nice! Seems like a good use case for fuzz testing.

core/test/libraries/Rewards.t.sol
```diff
@@ -0,0 +1,305 @@
...
+        assertEq(pool.rewards(a), 0);
+        assertEq(pool.rewards(b), 0);
+
+        pool.setRate(rate);
+        pool.mint(a, shares);
+        skip(60);
+
+        console.log(pool.balanceOf(a));
+        console.log(pool.rewards(a));
+        console.log(shares, rate);
```
aneesh1: Accident?

core/test/libraries/Log2.t.sol
```diff
@@ -0,0 +1,162 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.17;
+
+import "forge-std/Test.sol";
+
+import {Math} from "openzeppelin-contracts/utils/math/Math.sol";
+
+import "src/libraries/Log2.sol";
+
+contract Log2Test is Test {
```
aneesh1, line 10: Let me know if I missed anything or if we have another test that would highlight this behavior, but an idea I have for a fuzz test would be something like
`assertEq(log2(exp2(input)), input)`. IMO this would help highlight the packing / unpacking into log space. Although a test for a function that calls [`_accumulate`](https://github.com/aloelabs/aloe-ii/pull/94/files#diff-83c7700799ccedfab3e92b77e10391f9f2590d1dbe577c20eedb34e1f1d9e64bR137) would *technically* do the same thing (which I think we have), I think an explicit test would also be a good idea.

aneesh1 (approved):
Rewards are dependent upon the total supply. It makes sense that:
- [x] Any functions that result in a change in the total supply (like `_mint` or `_redeem`) AND change the amount of token that a user has should call `updatePoolState` and `updateUserState` (`_mint_`, `_burn`)
- [x] Any functions that result in a change in the amount of token a particular user has should call `updateUserState` (`_transfer`)

core/src/libraries/Rewards.sol
```diff
@@ -0,0 +1,148 @@
...
+    function updateUserState(Storage storage store, uint144 accumulated, address user, uint256 balance) internal {
+        store.userStates[user] = previewUserState(store, accumulated, user, balance);
+    }
+
+    function previewPoolState(
+        Storage storage store,
+        uint144 accumulated,
+        uint256 totalSupply
+    ) internal view returns (PoolState memory poolState) {
+        unchecked {
```
IanWoodard, line 103: No need to have unchecked here as there is no math

IanWoodard (approved):
Looks good

core/test/libraries/Rewards.t.sol
haydenshively: Yes good catch

core/test/libraries/Log2.t.sol
haydenshively: Added ✅

core/src/libraries/Rewards.sol
haydenshively: Ack, but I do kinda like how it makes the `previewPoolState` logic match up with the `previewUserState` logic (same type of stuff being done, at the same level of indentation). Let's leave this in and see if auditors notice / care.

PR #95, opened 2023-06-14
Update invariants for courier edge cases

haydenshively:
While implementing rewards, I happened to turn up the depth on our invariant testing and came across a few bugs. Fixed here.

Main thing is in the `LenderHarness.redeem` function, where I had forgotten to account for the fact that (if there's a courier involved) more than just the requested `shares` will be redeemed. In fact, `shares` will be burned, and an additional `fee` will be transferred to the courier, leaving the user with a net change of `shares + fee`.


core/test/invariants/LenderHarness.sol
```diff
@@ -277,12 +281,15 @@ contract LenderHarness {
             require(LENDER.asset().balanceOf(recipient) == assetsBefore, "redeem: bad self reference");
         }
         if (owner != LENDER.RESERVE()) {
-            require(LENDER.balanceOf(owner) == sharesBefore - shares, "redeem: burn issue");
-            require(reservesAfter == reservesBefore + newReserves, "redeem: reserves issue");
+            require(LENDER.balanceOf(owner) == sharesBefore - shares - fee, "redeem: burn issue");
```
aneesh1, line 284: This demonstrates the `shares + fee` net change 👍

aneesh1 (approved):
LGTM!

PR #96, opened 2023-06-14
Remove nominalAssets in favor of nominalShares

haydenshively:
Since the logic in `Lender._burn` matches `_nominalShares` (not `_nominalAssets`) I figured it would make sense to remove `_nominalAssets`. If Solidity math were perfect, we wouldn't have to worry about this. But because of rounding, we want the logic to match _exactly_ wherever possible.

This does make certain functions (like `underlyingBalance`) gassier, but we never call them onchain, so it doesn't matter much.


core/src/Ledger.sol
```diff
@@ -170,17 +170,15 @@ contract Ledger {
...
-            return
-                _nominalAssets({
-                    account: account,
-                    inventory: lastBalance + (uint256(borrowBase) * borrowIndex) / BORROWS_SCALER,
-                    totalSupply_: totalSupply
-                });
+            uint256 inventory = lastBalance + (uint256(borrowBase) * borrowIndex) / BORROWS_SCALER;
+            uint256 totalSupply_ = totalSupply;
+
+            return _convertToAssets(_nominalShares(account, inventory, totalSupply_), inventory, totalSupply_, false);
```
aneesh1, line 181: I'm sure this works, but why not pass `totalSupply` directly?

aneesh1 (approved):
Looks good!

core/src/Ledger.sol
haydenshively: We use `totalSupply_` twice on this line, so it's better to cache it in memory than to read from storage twice.

PR #97, opened 2023-06-14
Remove unnecessary 512 bit multiplication from accrueInterest

haydenshively:
`oldBorrows` is at most 145 bits, and `1e18` is 59.79 bits, so their product will certainly fit in 256 bits. No need for the gassier 512 bit multiplication function.


core/.gas-snapshot
```diff
@@ -22,5 +22,5 @@ LiquidatorGasTest:test_noCallbackTwoAssets() (gas: 52073)
 LiquidatorGasTest:test_noCallbackTwoAssetsAndUniswapPosition() (gas: 94964)
 LiquidatorGasTest:test_warn() (gas: 27875)
 LiquidatorGasTest:test_withCallbackAndSwap() (gas: 123133)
-VolatilityGasTest:test_consult() (gas: 13218)
+VolatilityGasTest:test_consult() (gas: 13142)
```
haydenshively, line 25: Really not sure why this changed too

aneesh1: Yeah, this is a bit weird considering that `test_consult` calls `oracle.update(pool)` and `oracle.consult(pool)` of which neither should be impacted by this change. The difference in gas seems negligible though.

aneesh1 (approved):
Nice!

PR #98, opened 2023-06-14
Explain why sums don't add up in comments

haydenshively:


core/src/Ledger.sol
```diff
@@ -168,11 +169,23 @@ contract Ledger {
         return balances[account] % Q112;
     }
 
+    /**
+     * @notice The amount of `asset` owed to `account`, i.e. the value that `maxWithdraw` would
+     * return if outstanding borrows weren't a constraint. Fees owed to couriers are automatically
+     * subtracted from this value in real-time, but couriers themselves won't receive earnings
+     * until users `redeem` or `withdraw`.
+     * @dev Because of the fees, ∑underlyingBalances != totalAssets
+     */
```
aneesh1: From the code it is clear that fees are included in the computation and the comment *implies* that fees are included:

> ```Fees owed to couriers are automatically subtracted from this value in real-time, but couriers themselves won't receive earnings```

Feel free to ignore, how about this for the first sentence:
```* @notice The amount of `asset` owed to `account`, after accruing the latest interest.```

This way it's explicit how `underlyingBalance` and `underlyingBalanceStored` are different.

File: core/src/Ledger.sol [hunk 2]
```diff
@@ -205,6 +218,11 @@ contract Ledger {
                            ERC4626 ACCOUNTING
     //////////////////////////////////////////////////////////////*/
 
+    /**
+     * @notice The total amount of `asset` under management
+     * @dev `convertToShares(totalAssets()) != totalSupply()` due to reserves inflation. If you need
+     * the up-to-date supply, use `stats()`
```
aneesh1, line 224: Good callout 👍

aneesh1 (approved):
LGTM

core/src/Ledger.sol
haydenshively: Gracias

PR #99, opened 2023-06-14
Store couriers in the Factory instead of individual Lenders

haydenshively:
This means that couriers can enroll just once, instead of having to enroll in every individual Lender.

Unfortunately had to rewrite the `setUp` functions of a lot of the tests so that the `Factory` storage is actually available (previously we were deploying raw `Lender`s sans any `Factory`).

Feel free to ignore test changes when reviewing.


core/.gas-snapshot
```diff
@@ -1,26 +1,26 @@
...
+BorrowerGasTest:test_repay() (gas: 103572)
+BorrowerGasTest:test_uniswapDepositInBorrower() (gas: 299971)
 BorrowerGasTest:test_uniswapDepositStandard() (gas: 167558)
-BorrowerGasTest:test_uniswapWithdraw() (gas: 136485)
-BorrowerGasTest:test_withdraw() (gas: 123651)
-FactoryGasTest:test_createBorrower() (gas: 156320)
-FactoryGasTest:test_createMarket() (gas: 3789361)
-LenderGasTest:test_accrueInterest() (gas: 46555)
+BorrowerGasTest:test_uniswapWithdraw() (gas: 140522)
+BorrowerGasTest:test_withdraw() (gas: 128504)
```
haydenshively, line 10: Increases in Borrower gas are due to the tests using proxies, as is done in prod. Previously the tests were dealing with a raw Borrower.

Jumping through proxy hoops = more gas.

PR #101, opened 2023-06-15
Re-ordering things and disabling extra lock

### LINKED ISSUE #59: Ordering of actions within `deposit`/`redeem`/`borrow`/`repay` could be more uniform
Have a few common themes, such as:
- loading from storage
- converting shares --> assets or amounts --> units (and vice versa)
- verifying tokens were sent / sending tokens
- minting / burning shares

Make sure these are ordered consistently across actions, following checks-effects-interactions whenever possible.

haydenshively:
- Re-order methods in Ledger to match Lender (1. deposit, 2. mint, 3. redeem, 4. withdraw)
- Re-order actions within deposit/redeem/borrow/repay to follow the "checks, effects, interactions" pattern. Closes #59. _In theory (but please check me on this)_ this means we no longer need to turn on the reentrancy lock in most cases. Previously we locked at the beginning of every function, and unlocked at the end. Now we only lock in the `flash` function.


core/src/Lender.sol
```diff
@@ -104,26 +104,28 @@ contract Lender is Ledger {
...
+        // Save state to storage (thus far, only mappings have been updated, so we must address everything else)
+        _save(cache, /* didChangeBorrowBase: */ false);
+
         // Ensure tokens are transferred
         ERC20 asset_ = asset();
-        cache.lastBalance += amount;
         bool didPrepay = cache.lastBalance <= asset_.balanceOf(address(this));
         if (!didPrepay) {
             asset_.safeTransferFrom(msg.sender, address(this), amount);
         }
```
haydenshively, line 127: These are the only calls in `deposit` that call outside of the Aloe protocol. In fact, the `safeTransferFrom` call is the only non-view function that even calls outside of this _contract_.

That means it's classified as "interactions" in checks/effects/interactions, so it should go last.

core/src/Lender.sol
```diff
@@ -133,27 +135,29 @@ contract Lender is Ledger {
...
-        // Guard against reentrancy, accrue interest, and update reserves
-        (Cache memory cache, uint256 inventory) = _load();
-
-        amount = _convertToAssets(shares, inventory, cache.totalSupply, /* roundUp: */ false);
-        require(amount != 0, "Aloe: zero impact");
-
         if (msg.sender != owner) {
             uint256 allowed = allowance[owner][msg.sender];
             if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares;
         }
```
haydenshively, line 141: This is a "check" that we can perform without loading or computing anything else, so it goes first.

core/src/Lender.sol
```diff
@@ -133,27 +135,29 @@ contract Lender is Ledger {
...
-        // Transfer tokens
+        // Assume tokens are transferred
         cache.lastBalance -= amount;
-        asset().safeTransfer(recipient, amount);
 
         // Save state to storage (thus far, only mappings have been updated, so we must address everything else)
         _save(cache, /* didChangeBorrowBase: */ false);
 
+        // Transfer tokens
+        asset().safeTransfer(recipient, amount);
```
haydenshively, line 159: Similar to `deposit`, this is the only call in `redeem` that leaves the protocol. Interaction --> goes last.

core/src/Lender.sol
```diff
@@ -456,7 +467,6 @@ contract Lender is Ledger {
         cache = Cache(totalSupply, lastBalance, lastAccrualTime, borrowBase, borrowIndex);
         // Guard against reentrancy
         require(cache.lastAccrualTime != 0, "Aloe: locked");
-        lastAccrualTime = 0;
```
haydenshively, line 459: We no longer lock every time we `_load` (pun intended)

core/src/Lender.sol
```diff
@@ -477,6 +487,6 @@ contract Lender is Ledger {
 
         totalSupply = cache.totalSupply.safeCastTo112();
         lastBalance = cache.lastBalance.safeCastTo112();
-        lastAccrualTime = uint32(block.timestamp); // Disables reentrancy guard
+        lastAccrualTime = uint32(block.timestamp); // Disables reentrancy guard if there was one
```
IanWoodard, line 490: Why keep this then?

IanWoodard (approved):

core/src/Lender.sol
haydenshively: Even if we don't care about the reentrancy guard at all, we still want to update `lastAccrualTime` to the current block's timestamp, because we did, in fact, accrue interest (as part of `_load`)

PR #102, opened 2023-06-17
Constrain yield per second so that rateModel upgrades are safe

haydenshively:
Since we plan for `rateModel` to be upgradeable, it's critical that we place safeguards around calls to it -- essentially input validation. This change moves the yieldPerSecond --> accrualFactor math _out of the RateModel_ and into a library. Though the library is in the `RateModel.sol` file, its bytecode is actually included as part of `Lender`, and therefore _is not_ upgradeable.

For this reason, it's safe to put our safeguards in the library. It checks the following:
- the `rateModel` can't use more than 10000 gas (no griefing)
- if the `rateModel` reverts, it sets the rate to 0. i.e. no interest accrues (but that's way better than bricking the contract!)
- if the `rateModel` returns a rate higher than `MAX_RATE`, it gets clipped (people have at least 1 year to withdraw before a nefarious or poorly implemented `rateModel` could cause any overflows)


core/src/RateModel.sol
```diff
@@ -3,27 +3,59 @@ pragma solidity 0.8.17;
```
haydenshively, line 3: Idk why the git history on this file is so ugly. If you just open the old version and new version in separate windows it's much easier to see what's going on.

core/src/RateModel.sol
```diff
@@ -3,27 +3,59 @@ pragma solidity 0.8.17;
...
+
 /// @title RateModel
 /// @author Aloe Labs, Inc.
 /// @dev "Test everything; hold fast what is good." - 1 Thessalonians 5:21
-contract RateModel {
+contract RateModel is IRateModel {
     uint256 private constant A = 6.1010463348e20;
 
-    uint256 private constant B = 1e12 - A / 1e18;
+    uint256 private constant B = A / 1e18;
```
aneesh1, line 25: Why isn't `B` still `1e12 - A / 1e18`?

core/src/RateModel.sol
```diff
@@ -3,27 +3,59 @@ pragma solidity 0.8.17;
...
 
-    function getAccrualFactor(uint256 elapsedTime, uint256 utilization) external pure returns (uint256) {
+    /// @inheritdoc IRateModel
+    function getYieldPerSecond(uint256 utilization, address) external pure returns (uint256) {
         unchecked {
-            uint256 rate = computeYieldPerSecond(utilization);
+            return (utilization < 0.99e18) ? A / (1e18 - utilization) - B : 60400;
```
aneesh1, line 30: Looks different than what we previously had: 
```
return (utilization < 0.99e18) ? B + A / (1e18 - utilization) : 1000000060400;
```
Why?

core/src/RateModel.sol
```diff
@@ -3,27 +3,59 @@ pragma solidity 0.8.17;
...
+
+library SafeRateLib {
+    using FixedPointMathLib for uint256;
 
-            if (elapsedTime > 1 weeks) elapsedTime = 1 weeks;
+    function getAccrualFactor(IRateModel rateModel, uint256 utilization, uint256 dt) internal view returns (uint256) {
+        uint256 rate;
 
-            return FixedPointMathLib.rpow(rate, elapsedTime, 1e12);
+        // Essentially `rate = rateModel.getYieldPerSecond(utilization, address(this)) ?? 0`, i.e. if the call
```
aneesh1, line 41: Thanks for the explanation 👍

core/test/invariants/LenderInvariants.t.sol
aneesh1, line 1: Tests look good! Disregard if we already have this OR if this is unnecessary, but is it worth adding a test to highlight the behavior detailed in [this comment](https://github.com/aloelabs/aloe-ii/pull/102/files#diff-bb311647b01ab28c90b40798e954d391f1c20bf9e61607972c4f83557f9eaaa6R11-R21)? Essentially, the rate is `MAX_RATE` for 1 year and then `borrowIndex` overflows? Mentioning this here b/c this feels like an invariant.

core/test/Lender.t.sol
```diff
@@ -91,7 +98,7 @@ contract LenderTest is Test {
             18
         );
         // Make sure `borrowIndex` is just as precise as `accrualFactor`
```
aneesh1: `accuralFactor` should be `yieldPerSecond`?

core/.gas-snapshot
aneesh1, line 1: > the rateModel can't use more than 10000 gas (no griefing) 

How do we know that we're meeting this requirement?

aneesh1 (approved):
Looks pretty good!

core/src/RateModel.sol
haydenshively: Sorry, I should've mentioned this in the PR description. Previously, `computeYieldPerSecond` returned $1+x$ with $x$ being the equivalent of APY, denominated in seconds. Now it just returned $x$ (i.e. dropping the $1 +$ part).

Before
$B + A / (1e18 - utilization) = 1e12 - (A / 1e18) + A / (1e18 - utilization) = A / (1e18 - utilization) - A / 1e18 + 1e12$

After (we _don't_ want the extract $1 +$ term, so subtract $1e12$
$A / (1e18 - utilization) - A / 1e18 + 1e12 - 1e12 = A / (1e18 - utilization) - A / 1e18$

core/src/RateModel.sol
haydenshively: See [above](https://github.com/aloelabs/aloe-ii/pull/102#discussion_r1234373457)

core/.gas-snapshot
haydenshively: Right now, we know we're meeting it because if we weren't, no interest would accrue, and many tests would fail.

In the future, if we write a more complicated `IRateModel`, we'll have to be careful to design it such that it meets this constraint.

core/test/Lender.t.sol
haydenshively: Yes that, and also I need to fix this as `yieldPerSecond` is _never_ greater than `1e12`

Good catch

core/test/invariants/LenderInvariants.t.sol
haydenshively: Good idea. Will add

PR #103, opened 2023-06-18
Fix one more thing in Lender invariants

haydenshively:


core/test/invariants/LenderHarness.sol
```diff
@@ -532,6 +535,15 @@ contract LenderHarness {
...
     }
 
+    function redeemWithLenderAsAssetReceiverAndCourier(uint112 shares, uint16 i) external returns (uint256 amount) {
+        if (LENDER.courierOf(address(0)) != 0) {
+            amount = redeem(shares, address(LENDER), address(0));
+        } else {
+            uint256 count = courierIds.length;
+            if (count != 0) creditCourier(courierIds[i % count], address(0));
+        }
+    }
```
haydenshively, line 545: This additional "special case" test helped me identify the issue with the `redeem` function -- without it it took dozens of runs of the test suite to see errors, with it the errors showed up like every time.

Keeping so that any regressions will show up quickly.

PR #104, opened 2023-06-20
Prevent couriers from claiming rewards

haydenshively:


core/src/Factory.sol
```diff
@@ -111,19 +118,48 @@ contract Factory {
...
+        unchecked {
+            uint256 count = lenders.length;
+            for (uint256 i = 0; i < count; i++) {
+                earned += lenders[i].claimRewards(msg.sender);
+            }
+        }
 
-        emit EnrollCourier(id, wallet, cut);
+        REWARDS_TOKEN.safeTransfer(beneficiary, earned);
     }
```
haydenshively, line 164: It wasn't strictly necessary that I move this here, but it's convenient because it means the `isCourier` mapping can be accessed locally, and also lets users claim rewards from multiple `Lender`s at once.

aneesh1 (approved):
LGTM

PR #106, opened 2023-06-22
Include courier crediting in deposit function

haydenshively:
It was really annoying to have to call a separate function (`creditCourier`) before depositing. This fixes that, although it does add a bit of complexity.


core/test/LenderReferrals.t.sol
aneesh1, line 1: Tests look good! Two suggestions:
1. Should we change the name of this file from `LenderReferrals.t.sol` to `LenderRewards.t.sol`? 
2. Do we have a test where we `deposit` into an account with 0 balance along with a courier? Based on my understanding of `_mint`, this would set the courier accordingly.

aneesh1 (approved):
Streamlines the whole process - nice!

core/test/LenderReferrals.t.sol
haydenshively: 1. It's still primarily testing referrals, rewards are separate (though everything does kinda happen inside of `_mint` now)
2. Yes! [Here](https://github.com/aloelabs/aloe-ii/blob/a84d1501ddce45d611fc34966544025a632645fd/core/test/LenderReferrals.t.sol#L100) and also the invariants cover this

core/test/gas/LenderGas.t.sol
```diff
@@ -59,7 +59,7 @@ contract LenderGasTest is Test {
         vm.prank(alice);
         asset.approve(address(this), type(uint256).max);
 
-        // `alice` allows this test contract to manage his WETH+
+        // `alice` allows this test contract to manage her WETH+
```
IanWoodard, line 62: nice catch 😂

core/src/RateModel.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity 0.8.17;
 
 import {FixedPointMathLib} from "solmate/utils/FixedPointMathLib.sol";
 
-import {MAX_RATE, ONE} from "./libraries/constants/Constants.sol";
+import {MAX_RATE, ONE} from "src/libraries/constants/Constants.sol";
```
IanWoodard: Why the change?

IanWoodard (approved):

core/src/RateModel.sol
haydenshively: Good catch. Messed up the merge.

PR #112, opened 2023-07-12
Let owner choose recipient in uniswapWithdraw

haydenshively:
I believe this is safe to do because it's a shortcut to something that was already possible, namely:
- burn & collect a Uniswap position, with all funds ending up in your `Borrower`
- configure `allowances` so that you can `transferFrom` those funds from your `Borrower` to a place of your choosing

becomes:
- burn & collect a Uniswap position, with all funds ending up at a place of your choosing

This change has a negligible impact on gas tests in `core`, and enabled me to make the `periphery`'s `UniswapNFTManager` slightly more gas-efficient.


periphery/src/managers/UniswapNFTManager.sol
```diff
@@ -36,10 +36,7 @@ contract UniswapNFTManager is IManager {
         // move position from NonfungiblePositionManager to Borrower
         if (liquidity < 0) {
             // safety checks since this contract will be approved to manager users' positions
-            {
-                require(FACTORY.isBorrower(msg.sender));
-                require(owner == NFT_MANAGER.ownerOf(tokenId));
-            }
+            require(FACTORY.isBorrower(msg.sender) && owner == NFT_MANAGER.ownerOf(tokenId));
```
haydenshively, line 39: No reason for this to have been split into 2 requires instead of 1. Fixed.

periphery/src/managers/UniswapNFTManager.sol
```diff
@@ -49,9 +46,12 @@ contract UniswapNFTManager is IManager {
...
             ERC20 token1 = borrower.TOKEN1();
 
-            (uint256 burned0, uint256 burned1, , ) = borrower.uniswapWithdraw(lower, upper, uint128(liquidity));
-            token0.safeTransferFrom(msg.sender, address(this), burned0);
-            token1.safeTransferFrom(msg.sender, address(this), burned1);
+            (uint256 burned0, uint256 burned1, uint256 collected0, uint256 collected1) = borrower.uniswapWithdraw(
+                lower,
+                upper,
+                uint128(liquidity),
+                address(this)
```
haydenshively, line 53: By specifying `address(this)` as the recipient, we avoid having to do a separate `safeTransferFrom` from the borrower to `address(this)`. This also means we don't need to set the `allowances` array when calling modify, which saves gas in and of itself.

periphery/src/managers/UniswapNFTManager.sol
```diff
@@ -65,6 +65,9 @@ contract UniswapNFTManager is IManager {
                     deadline: block.timestamp
                 })
             );
+
+            token0.safeTransfer(owner, collected0 - burned0);
+            token1.safeTransfer(owner, collected1 - burned1);
```
haydenshively, line 70: But now we need to send leftover funds back to the owner -- the burned amounts are used to re-create the Uniswap position, but fees (collected - burned) need to go somewhere. Previously they would've just sat in the `Borrower`, now we're sending them to the owner.

aneesh1 (approved):
looks good!

periphery/src/managers/UniswapNFTManager.sol
IanWoodard: Well, there is a reason... it just isn't a very good one

IanWoodard (approved):

PR #115, opened 2023-07-23
Detect Uniswap TWAP oracle manipulation and pause borrows

haydenshively:
In PoW, manipulating the Uniswap TWAP oracle was very costly because you were could not know ahead of time whether you would control transaction ordering for 2 blocks in a row. Therefore, if you move the price up/down in one block (to begin manipulating the TWAP) you may not have a chance to move it back down/up in the next block. If you fail to do so, arbitrageurs will do it instead, and you'll lose a ton of money.

In PoS, proposers for the epoch (32 blocks, 6.4 minutes) are known ahead of time. If you control 2 or more blocks in a row, you can manipulate the TWAP without fear of arbitrageurs, meaning that the only cost is the swap fee (pay once on the way up/down, and once on the way down/up). Furthermore, for large stakers like Lido, the odds of getting 2 or more blocks in a row are quite good.

See [here](https://blog.uniswap.org/uniswap-v3-oracles) for more info.

Because Aloe uses the TWAP oracle, we need a way to detect whether it's being manipulated. Our manipulation metric is defined as follows:

$`\begin{align}
&TWAP_{measured, -60mins \rightarrow now} - TWAP_{measured, -60mins \rightarrow -30mins} = \\
&= \frac{1}{2} \left( TWAP_{true, -30mins \rightarrow now} - TWAP_{true, -60mins \rightarrow -30mins} \right) \\
&+ \frac{1}{60mins} \left( \sum_{-30mins}^{now}(tick_{manipulated} - tick_{true})dt - \sum_{-60mins}^{-30mins}(tick_{manipulated} - tick_{true})dt \right)
\end{align}`$

We assume that the amount of manipulation required to create bad debt >>> the true price change in an hour, such that $\left( TWAP_{true, -30mins \rightarrow now} - TWAP_{true, -60mins \rightarrow -30mins} \right) \approx 0$. This leaves us with the last term of the equation, or in english:

"The total amount of manipulation in this window, minus the total amount of manipulation in the previous window" (a window is 30 minutes)

We calculate this value and watch it at all times. If it ever rises above a certain threshold, we pause certain activities for the following 30 minutes. Namely, all calls to `modify` must result in 0 debt, and by extension no new debt can be created.

[This Desmos](https://www.desmos.com/calculator/wylksvom1g) shows how the manipulation metric behaves when the true price is constant (drift=0, volatility=0), and an attacker manipulates the price in 2 distinct chunks to raise the TWAP by 10%. In the real world, prices aren't constant and attackers may manipulate things at more than 2 times. To capture this, I simulated prices as GBM (Geometric Brownian Motion) and generated random attacks that lift the TWAP enough to create bad debt in the protocol.

The X and Y axis are the volatility and drift inputs for the GBM. The Z axis is the value of our manipulation metric as defined above. Blue points are the non-manipulated GBM trajectories, and colored points are manipulated ones. Note that as volatility increases, collateral factors decrease, so the _amount of manipulation required to create bad debt_ increases. This explains the exponential increase in our manipulation metric. It levels out at extreme volatilities because the protocol has a hard cap (see `MAX_SIGMA` in [Constants.sol](https://github.com/aloelabs/aloe-ii/blob/c3822e7b6d0f62c95486d5bc46017c959ad2a2ee/core/src/libraries/constants/Constants.sol#L37)).

Other notes: The color of the non-blue points answers the following question: "For all price trajectories and attacks simulated at this {x, y}, what was the _minimum_ number of blocks for which our heuristic indicated manipulation?". In other words, yellow points mean there was plenty of time to recognize the attack and pause the protocol, while purple points mean it was harder to recognize. In the worst case there'd be ~50 blocks in which someone could call `modify` and trigger a pause.

manipulation_metric

Finally, to see why I set the threshold the way I did, this next graph may be informative. Instead of having the Z axis show the raw value of our manipulation metric, it now shows $\frac{manipulationMetric}{threshold}$. This makes it easier to see that I set the threshold to be 3x smaller than the sneakiest manipulation across the board. i.e. the green threshold surface is at 1, and the lowest yellow points congregate around 3. 3x feels like enough of a safety factor, and any more than that (4x, 5x, etc.) starts to generate false positives -- some blue points end up above the green surface.

manipulation_metric2


core/src/libraries/Oracle.sol
```diff
@@ -13,32 +15,74 @@ library Oracle {
...
+            //
+            //                                      ≈ (sumManip0ToW - sumManipWTo2W) / (2T)
+            //
+            // The TWAP we care about (see a few lines down) is measured over the interval [-w, 0). The result we've
+            // just derived contains `sumManip0ToW`, which is the sum of all manipulation in that same interval. As
+            // such, we use it as a metric for detecting manipulation. NOTE: If an attacker manipulates things to
+            // the same extent in the prior interval [-2w, -w), the metric will be 0. To guard against this, we must
+            // to watch the metric over the entire window. Even though it may be 0 *now*, it will have risen past a
+            // threshold at *some point* in the past `UNISWAP_AVG_WINDOW` seconds.
+            metric = uint56(SoladyMath.dist(meanTick0To2W, meanTickWTo2W));
```
haydenshively, line 76: $|TWAP_{measured, -60mins \rightarrow now} - TWAP_{measured, -60mins \rightarrow -30mins}|$

core/src/libraries/Oracle.sol
```diff
@@ -13,32 +15,74 @@ library Oracle {
...
 
-            sqrtMeanPriceX96 = TickMath.getSqrtRatioAtTick(arithmeticMeanTick);
-            secondsPerLiquidityX128 = secondsPerLiquidityCumulativeX128s[1] - secondsPerLiquidityCumulativeX128s[0];
+            // Compute arithmetic mean tick over `UNISWAP_AVG_WINDOW`, always rounding down to -inf
+            int256 delta = tickCumulatives[2] - tickCumulatives[1];
+            int256 meanTick0ToW = delta / int32(UNISWAP_AVG_WINDOW);
```
haydenshively, line 80: $TWAP_{measured, -30mins \rightarrow now}$

This is the one we actually use as a price. Everything else is to determine whether this is manipulated or not.

.solhint.json
```diff
@@ -4,7 +4,7 @@
   "rules": {
     "prettier/prettier": "error",
 
-    "code-complexity": ["error", 6],
+    "code-complexity": ["error", 7],
```
IanWoodard, line 7: 😨

core/src/Borrower.sol
```diff
@@ -272,16 +272,21 @@ contract Borrower is IUniswapV3MintCallback {
...
         Assets memory assets = _getAssets(positions_, prices, false);
         (uint256 liabilities0, uint256 liabilities1) = _getLiabilities();
 
         require(BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: unhealthy");
         unchecked {
-            if (liabilities0 + liabilities1 > 0) require(address(this).balance > ante, "Aloe: missing ante");
+            if (liabilities0 + liabilities1 > 0)
+                require(
+                    address(this).balance > ante && !isSus && block.timestamp > pausedUntilTime,
+                    "Aloe: ante or TWAP"
```
IanWoodard: not a very useful message

core/test.sh
```diff
@@ -21,20 +21,6 @@ if [ -z "$CI" ]; then
...
-      n=$(grep -rnw --include=\*.sol '.' -e "$constant" | wc -l)
-      if (( n > 1)); then
-        echo "❌ `${constant}` is defined more than once" >> $GITHUB_STEP_SUMMARY
-        exit 1
-      fi
-    done <<< "$constants"
-
-    echo "✅ Each constant is only defined once" >> $GITHUB_STEP_SUMMARY
-fi
-
```
IanWoodard: Why are we removing this in this PR?

core/src/libraries/constants/Constants.sol
```diff
@@ -70,3 +70,13 @@ uint256 constant FEE_GROWTH_SAMPLE_PERIOD = 5 minutes;
...
+
+/// @dev Assume someone is manipulating the Uniswap TWAP oracle. To steal money from the protocol and create bad debt,
+/// they would need to change the TWAP by a factor of (1 / collateralFactor), where the collateralFactor is a function
+/// of volatility. We have a manipulation metric that increases as an attacker tries to change the TWAP. If this metric
+/// rises above a certain threshold, certain functionality will be paused, e.g. no new debt can be created. The
+/// threshold is calculated as follows:
+///
+///   manipulationThreshold = 2 * log_1.0001(1 / collateralFactor) / MANIPULATION_THRESHOLD_DIVISOR
+///
+uint24 constant MANIPULATION_THRESHOLD_DIVISOR = 24;
```
IanWoodard, line 82: Is this arbitrary?

core/src/libraries/BalanceSheet.sol
```diff
@@ -119,4 +119,18 @@ library BalanceSheet {
             }
         }
     }
+
+    /// @dev Equivalent to log_{1.0001}(1e18 / cf) / 12 assuming `MANIPULATION_THRESHOLD_DIVISOR = 24`
+    function _manipulationThreshold(uint256 cf) private pure returns (uint24) {
+        return uint24(-TickMath.getTickAtSqrtRatio(uint160(cf)) - 501937) / MANIPULATION_THRESHOLD_DIVISOR;
```
IanWoodard, line 125: What is 501937?

IanWoodard (approved):

.solhint.json
haydenshively: I know :(

core/src/Borrower.sol
haydenshively: I know, but gotta keep it under 32 characters for gas efficiency. Willing to change if you have a suggestion?

core/test.sh
haydenshively: The comment that I wrote for the `MANIPULATION_THRESHOLD_DIVISOR` constant was triggering weird false positives. I think it has something to do with `grep` matching on the `=` even though the `=` is inside the comment.

If you know how to fix it (instead of deleting) lmk.

core/src/libraries/constants/Constants.sol
haydenshively: No, based on simulation.

core/src/libraries/BalanceSheet.sol
haydenshively: Our goal is to calculate $log_{1.0001}$ of the reciprocal of the collateral factor. But instead of writing a new function that calculates $log_{1.0001}$ directly, we want to use the existing `TickMath.getTickAtSqrtRatio(x)`.

Let $F(x)$ represent `TickMath.getTickAtSqrtRatio(x)`

$F(2^{96} * \sqrt{x}) = log_{1.0001}x$

Let $y = \frac{x^2}{2^{192}}$. Then we have:

$F(2^{96} * \sqrt{y}) = log_{1.0001}y$
$F(2^{96} * \sqrt{\frac{x^2}{2^{192}}}) = log_{1.0001}(\frac{x^2}{2^{192}})$
$F(2^{96} * \frac{x}{2^{96}}) = 2log_{1.0001}(x) - 192log_{1.0001}(2)$
$F(x) = 2log_{1.0001}(x) - 192log_{1.0001}(2)$

Rearranging, we get:

$log_{1.0001}(x) = \frac{1}{2} \cdot (F(x) + 192log_{1.0001}(2))$
$log_{1.0001}(x) = \frac{1}{2} \cdot (F(x) + 1330909.128)$

Then we deal with the extra factor of $10^{18}$ in our collateral factor.

$log_{1.0001}(c * 10^{18}) = log_{1.0001}(c) + 18log_{1.0001}(10)$
$log_{1.0001}(c * 10^{18}) = log_{1.0001}(c) + 414486$
$log_{1.0001}(c) = log_{1.0001}(c * 10^{18}) - 414486$

Plugging in the previous equation:

$log_{1.0001}(c) = \frac{1}{2} \cdot (F(c * 10^{18}) + 1330909.128) - 414486.04$
$log_{1.0001}(c) = \frac{1}{2} \cdot (F(c * 10^{18}) + 1330909.128 - 828972.07)$
$log_{1.0001}(c) = \frac{1}{2} \cdot (F(c * 10^{18}) + 501937.06)$

Finally, taking the reciprocal is the same as multiplying by -1 in log space, so:

$log_{1.0001}(\frac{1}{c}) = \frac{1}{2} \cdot (-F(c * 10^{18}) - 501937.06)$

Translating to code and ignoring data types, this would be:

```Solidity
(-TickMath.getTickAtSqrtRatio(cf) - 501937) / 2;
```

This is essentially the base case for the manipulation threshold, so you can see the `MANIPULATION_THRESHOLD_DIVISOR` that I use in the actual code is just to scale things down and make our heuristic more sensitive.

haydenshively: TLDR:

$`501937 = 192log_{1.0001}(2) - 2*18*log_{1.0001}(10) = \frac{192log(2) - 36log(10)}{log(1.0001)}`$

core/src/Borrower.sol
```diff
@@ -272,16 +272,21 @@ contract Borrower is IUniswapV3MintCallback {
         if (allowances[0]) TOKEN0.safeApprove(address(callee), 1);
         if (allowances[1]) TOKEN1.safeApprove(address(callee), 1);
 
-        (uint256 ante, uint256 nSigma) = FACTORY.getParameters(UNISWAP_POOL);
+        (uint256 ante, uint256 nSigma, uint256 pausedUntilTime) = FACTORY.getParameters(UNISWAP_POOL);
 
-        Prices memory prices = _getPrices(nSigma);
+        (Prices memory prices, bool isSus) = _getPrices(nSigma);
```
aneesh1, line 277: 🤣

core/.gas-snapshot
aneesh1, line 1: Gas 📈 😮 , but it looks like some optimizations in https://github.com/aloelabs/aloe-ii/pull/119 _may_ have reduced it. Not 100% confident, but briefly skimmed over the diff and it looks some commits from this branch are in there, so not a cause for concern.

core/src/Factory.sol
```diff
@@ -80,6 +81,13 @@ contract Factory {
         lenderImplementation = address(new Lender(address(this)));
     }
 
+    function pause(IUniswapV3Pool pool) external {
+        require(isBorrower[msg.sender]);
+        unchecked {
+            getParameters[pool].pausedUntilTime = uint32(block.timestamp) + UNISWAP_AVG_WINDOW;
+        }
+    }
```
aneesh1, line 89: Correct me if I'm wrong, but it seems like a borrower could call `FACTORY.pause()` anytime. I'm considering a case when `FACTORY.pause()` is in the callback function that is invoked in `modify`. If this is not possible, feel free to ignore, but should we only permit `pause` when something `isSus`?

core/test/libraries/TickMath.t.sol
```diff
@@ -66,26 +66,89 @@ contract TickMathTest is Test {
...
+            getSqrtRatioAtTickOriginal(tick);
+            tick = TickMath.MAX_TICK;
+        }
+
+        uint256 a = TickMath.getSqrtRatioAtTick(int24(tick));
+        uint256 b = getSqrtRatioAtTickOriginal(int24(tick));
 
         assertEq(a, b);
     }
 }
```
aneesh1: 👏 💯 Although your comments on the code for `getSqrtRatioAtTick` were pretty thorough, this test explicitly shows that the two are equal.

aneesh1 (approved):
Thanks for providing a thorough explanation with graphs and the [link to the Uniswap article ](https://blog.uniswap.org/uniswap-v3-oracles) - these resources provided me with enough context to review this change. 

I think we're computing the metric correctly and based on the explanation, it should be sufficient to detect price manipulation. I did have some comments about the increased gas that resulted from the changes in this PR and a question about the `pause` method. Apart from that, looks good!

core/.gas-snapshot
haydenshively: Yeah the increase in gas is a result of querying the oracle for an additional timestamp ({60 min, 30 min, now} instead of just {30 min, now}). Each timestamp is an extra binary search in the oracle code, so it's painful.

#119 does indeed fix this.

core/src/Factory.sol
haydenshively: Good callout. I hadn't yet considered this, but I've double checked and I think we're fine. Nowhere in `Borrower` do we allow arbitrary calls to arbitrary addresses, even in `modify`. i.e. the owner can pass in any `callee` they want, but we only do `callee.callback` _not_ `callee.pause`.

PR #116, opened 2023-07-24
Use SoladyMath wherever possible

haydenshively:


haydenshively:
Doesn't save any gas (probably because we have optimizations turned all the way up and --via-ir) and introduces extra dependencies.

PR #117, opened 2023-07-26
Create alternative to Uniswap's observe method, no binary search

haydenshively:
Just creating the method. Will integrate it with our stuff in the next PR.


core/test/libraries/Oracle.t.sol
```diff
@@ -0,0 +1,349 @@
...
+            liquidity
+        );
+
+        // Make sure they're the same
+        assertEq(tickCumB, tickCumA);
+        assertEq(liqCumB, liqCumA);
+    }
+
+    /// forge-config: default.fuzz.runs = 1024
+    function test_comparative_observeOld(uint32 secondsAgo, uint256 seed, bool initializing) public {
```
aneesh1, line 78: Did you mean to add `assertEq`'s to this test? In the other `_comparative_` tests you had an `assertEq` comparing two values - that's why I'm curious?

core/src/libraries/Oracle.sol
```diff
@@ -86,6 +86,75 @@ library Oracle {
...
+
+                if (timeL < target && target < timeR) {
+                    uint56 delta = uint56(target - timeL);
+                    uint56 denom = uint56(timeR - timeL);
+                    // Uniswap divides before multiplying, so we do too
+                    return (
+                        tickCumL + ((tickCumR - tickCumL) / int56(denom)) * int56(delta),
+                        liqCumL + uint160(((liqCumR - liqCumL) * delta) / denom)
+                    );
+                }
```
aneesh1, line 149: Linear interpolation for target that exists between two observations - looks correct ✅

core/src/libraries/Oracle.sol
```diff
@@ -86,6 +86,75 @@ library Oracle {
...
+            for (uint256 i = 0; i < observationCardinality; i++) {                
+                if (timeL == target) {
+                    return (tickCumL, liqCumL);
+                }
+
+                if (timeL < target && seed == observationIndex) {
+                    uint56 delta = uint56(target - timeL);
+                    return (
+                        tickCumL + tick * int56(delta),
+                        liqCumL + (uint160(delta) << 128) / (liquidity > 0 ? liquidity : 1)
```
aneesh1, line 134: Why the `<<128`? I understand what you're doing in the line above, but this one I'm not so sure. Given that your `_comparative_` tests pass though, it must be right, but I don't see why you need this.

aneesh1 (approved):
Looks good! Two questions, but b/c you have unit tests I'm convinced the code works. Side note: any benefits to this approach? Less gas perhaps?

core/test/libraries/Oracle.t.sol
haydenshively: In this case I'm expecting both methods to revert with code "OLD", instead of expecting both methods to have the same output.

core/src/libraries/BalanceSheet.sol
```diff
@@ -120,7 +120,7 @@ library BalanceSheet {
         }
     }
 
-    /// @dev Equivalent to log_{1.0001}(1e18 / cf) / 12 assuming `MANIPULATION_THRESHOLD_DIVISOR = 24`
+    /// @dev Equivalent to log_{1.0001}(1e18 / cf) / 12 assuming `MANIPULATION_THRESHOLD_DIVISOR` is 24
```
IanWoodard, line 123: 💯

core/src/libraries/Oracle.sol
```diff
@@ -86,6 +86,75 @@ library Oracle {
...
+                    return (
+                        tickCumL + ((tickCumR - tickCumL) / int56(denom)) * int56(delta),
+                        liqCumL + uint160(((liqCumR - liqCumL) * delta) / denom)
+                    );
+                }
+
+                (timeL, tickCumL, liqCumL) = (timeR, tickCumR, liqCumR);
+            }
+
+            revert("OLD");
```
IanWoodard, line 154: Very vague message

core/test/libraries/Oracle.t.sol
```diff
@@ -0,0 +1,349 @@
...
+        (, int24 tick, uint16 idx, uint16 cardinality, , , ) = POOL.slot0();
+        uint128 liquidity = POOL.liquidity();
+
+        (uint32 oldest, ) = _oldestAndNewest();
+        secondsAgo = uint32(bound(secondsAgo, block.timestamp - oldest + 1, block.timestamp));
+
+        // Uniswap method
+        uint32[] memory secondsAgos = new uint32[](1);
+        secondsAgos[0] = secondsAgo;
+        vm.expectRevert(bytes("OLD"));
```
IanWoodard, line 91: Is there a reason we hardcode all of these error messages in here rather than use a constant in the test file?

core/src/libraries/Oracle.sol
```diff
@@ -86,6 +86,75 @@ library Oracle {
...
+     */
+    function observe(
+        IUniswapV3Pool pool,
+        uint32 target,
+        uint256 seed,
+        int24 tick,
+        uint16 observationIndex,
+        uint16 observationCardinality,
+        uint128 liquidity
+    ) internal view returns (int56, uint160) {
```
IanWoodard, line 120: Are there any possible vulnerabilities we aren't considering that come with this approach?

IanWoodard (approved):

core/src/libraries/Oracle.sol
haydenshively: Chose this so that it matches Uniswap

core/test/libraries/Oracle.t.sol
haydenshively: Not particularly. Constant in the test file would work too but I don't mind find+replace 🤷‍♂️

core/src/libraries/Oracle.sol
haydenshively: The `liquidity` value from Uniswap has an extra factor of 2^128 in it, so when we're taking the reciprocal we need that same factor in the numerator.

core/src/libraries/Oracle.sol
haydenshively: I mean, unknown unknowns are unknown, but the tests show this method gives identical results to the method that Uniswap provides.

PR #119, opened 2023-07-27
Integrate oracle seeding into Borrower

haydenshively:
Gas savings!


core/src/libraries/Oracle.sol
```diff
@@ -5,36 +5,95 @@ import {FixedPointMathLib as SoladyMath} from "solady/utils/FixedPointMathLib.so
...
         unchecked {
+            int56[] memory tickCumulatives = new int56[](3);
+            uint160[] memory secondsPerLiquidityCumulativeX128s = new uint160[](3);
+
+            if ((seed >> 32) > 0) {
+                uint32[] memory secondsAgos = new uint32[](3);
+                secondsAgos[0] = UNISWAP_AVG_WINDOW * 2;
+                secondsAgos[1] = UNISWAP_AVG_WINDOW;
+                secondsAgos[2] = 0;
+                (tickCumulatives, secondsPerLiquidityCumulativeX128s) = pool.observe(secondsAgos);
```
haydenshively, line 56: If any of the highest 8 bits of `seed` are set, fallback to Uniswap's method that uses binary search (this is what we were using before).

core/src/libraries/Oracle.sol
```diff
@@ -129,6 +177,7 @@ library Oracle {
 
                 if (timeL < target && seed == observationIndex) {
                     uint56 delta = uint56(target - timeL);
+                    uint128 liquidity = pool.liquidity();
```
haydenshively, line 180: This is the only if statement where `liquidity` is necessary for the computation, so instead of fetching `liquidity` every time (and rarely using it) we fetch `liquidity` only after determining that it's needed.

PR #131, opened 2023-08-17
Fix liquidation conditionals to avoid zero liabilities swap

### LINKED ISSUE #68: Could sometimes return early in last 2 branches of `Borrower.liquidate`
After dividing by `strain`, if `liabilities0 == 0` (or `liabilities1 == 0`, depending on the branch) we could return early instead of passing execution to the liquidator.

Alternatively, just document it and make sure liquidators can handle this edge case.

haydenshively:
Fixes #68

### DIFF
--- a/core/src/Borrower.sol
+++ b/core/src/Borrower.sol
@@ -207,42 +207,46 @@ contract Borrower is IUniswapV3MintCallback {
             liabilities0 -= repayable0;
             liabilities1 -= repayable1;
 
-            if (liabilities0 + liabilities1 == 0 || (liabilities0 > 0 && liabilities1 > 0)) {
-                // If both are zero or neither is zero, there's nothing more to do.
-                // Callbacks/swaps won't help.
-                incentive1 = 0;
-            } else if (liabilities0 > 0) {
+            // Decide whether to swap or not
+            bool shouldSwap;
+            assembly ("memory-safe") {
+                // If both are zero or neither is zero, there's nothing more to do
+                shouldSwap := xor(gt(liabilities0, 0), gt(liabilities1, 0))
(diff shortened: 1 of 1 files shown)


core/src/Borrower.sol
```diff
@@ -207,42 +207,46 @@ contract Borrower is IUniswapV3MintCallback {
...
+            // Decide whether to swap or not
+            bool shouldSwap;
+            assembly ("memory-safe") {
+                // If both are zero or neither is zero, there's nothing more to do
+                shouldSwap := xor(gt(liabilities0, 0), gt(liabilities1, 0))
+                // Divide by `strain` and check again. This second check can generate false positives in cases
+                // where one division (not both) floors to 0, which is why we `and()` with the check above.
+                liabilities0 := div(liabilities0, strain)
+                liabilities1 := div(liabilities1, strain)
+                shouldSwap := and(shouldSwap, xor(gt(liabilities0, 0), gt(liabilities1, 0)))
```
aneesh1, line 219: 👍 If `liabilities0` or `liabilities1` gets truncated as a result of `/= strain`, then we won't swap. The liquidator would probably have to adjust the `strain` in order to perform the liquidation successfully.

core/src/Borrower.sol
```diff
@@ -207,42 +207,46 @@ contract Borrower is IUniswapV3MintCallback {
...
+                // If not swapping, set `incentive1 = 0`
+                incentive1 := mul(shouldSwap, incentive1)
```
aneesh1, line 221: TIL.

aneesh1 (approved):
I think this fixes the issue 👍

IanWoodard (approved):

PR #133, opened 2023-08-22
Add uniswapSwap sub-command to Borrower

haydenshively:
In the `BoostManager` contract I've been working on on master, we need to swap tokens. Without a `uniswapSwap` sub-command, I was forced to implement the swap in the manager contract itself. This isn't the end of the world, but felt slightly roundabout, so I decided to try moving the logic inside the `Borrower`.

Having done so, I'm not sure it's worth it. It will save ~50k (very rough estimate) when calls to `modify` happen to involve a trade on the Borrower's specific Uniswap pool. However, it increases gas required for other operations (see .gas-snapshot diff). And of course it's extra logic to worry about security-wise.

@aneesh1 and @IanWoodard what do you think? Should we include this sub-command or not?


haydenshively:
Closing since #134 addresses swap efficiency in a cleaner way.

PR #134, opened 2023-08-22
Remove allowances from Borrower.modify in favor of transfer sub-command

haydenshively:
## Explanation

Previously, a user would set `allowances` to `[true, true]` to enable arbitrary transfers out of their `Borrower`. Prior to the modify callback, we would set the allowance to `type(uint256).max`, and afterwards we would set it back to `1`. Note that we _did not_ set it to 0 -- this was intended to save gas on the next run, since it's cheaper to overwrite a non-zero value than it is to overwrite 0.

Unfortunately, Tether (USDT) does not support this flow (see line 205 on the [verified contract](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7#code)).

I realized we could fix this and simultaneously achieve the goals of #133 (without the mess of that PR). I simply replaced the allowances flow with a new `transfer` sub-command. This is:
- more compatible
- more secure
- more gas-efficient

Win win.

## Periphery Changes

- `WithdrawManager` is no longer necessary, as the same thing can be accomplished by encoding a call to `borrower.transfer` and sending it to the `SimpleManager`
- "transfer out" and "swap" actions are simplified in the `FrontendManager` since we no longer need to `transferFrom` the `Borrower`


core/src/Borrower.sol
```diff
@@ -370,6 +357,20 @@ contract Borrower is IUniswapV3MintCallback {
...
+     * @param amount0 The amount of `TOKEN0` to transfer
+     * @param amount1 The amount of `TOKEN1` to transfer
+     * @param recipient Receives the transferred tokens
+     */
+    function transfer(uint256 amount0, uint256 amount1, address recipient) external {
+        require(slot0.state == State.InModifyCallback);
+
+        if (amount0 > 0) TOKEN0.safeTransfer(recipient, amount0);
+        if (amount1 > 0) TOKEN1.safeTransfer(recipient, amount1);
+    }
```
aneesh1, line 372: Correct me if I'm wrong, but previously we had to `approve` the `callee` to spend some of this contract's tokens. Now, to achieve the same result, we'd have to transfer the tokens and the address of the recipient would be `address(callee)`?

periphery/src/managers/FrontendManager.sol
```diff
@@ -23,11 +23,9 @@ contract FrontendManager is IManager, IUniswapV3SwapCallback {
...
     }
 
-    /// @dev This is safe because though this contract will have many approvals, its assets are always transient.
-    // As long as we do `safeTransfer` and not `safeTransferFrom`, we're fine.
-    function uniswapV3SwapCallback(int256 amount0, int256 amount1, bytes calldata) external {
-        if (amount0 > 0) ERC20(IUniswapV3Pool(msg.sender).token0()).safeTransfer(msg.sender, uint256(amount0));
-        if (amount1 > 0) ERC20(IUniswapV3Pool(msg.sender).token1()).safeTransfer(msg.sender, uint256(amount1));
+    function uniswapV3SwapCallback(int256 amount0, int256 amount1, bytes calldata data) external {
+        Borrower borrower = abi.decode(data, (Borrower));
+        borrower.transfer(amount0 > 0 ? uint256(amount0) : 0, amount1 > 0 ? uint256(amount1) : 0, msg.sender);
```
aneesh1, line 28: Clean.

aneesh1 (approved):
LGTM

core/src/Borrower.sol
haydenshively: Yep!

periphery/src/managers/FrontendManager.sol
IanWoodard: Clean

IanWoodard (approved):

PR #139, opened 2023-08-29
Increase price precision from Q96 to Q128

### LINKED ISSUE #66: Additional safety check in `computeProbePrices`
When `sqrtMeanPriceX96` is near `TickMath.MIN_SQRT_RATIO` or `TickMath.MAX_SQRT_RATIO`, probe prices `a` and `b` may end up outside of TickMath's range. We never use the `getTickAtSqrtRatio` function, so that doesn't matter in and of itself. However, exceeding `type(uint160).max` could create problems when mulDiv'ing to get `priceX96`, depending on the data type used. Furthermore, if we _didn't_ have proper constraints on `MIN_SIGMA`, `MAX_SIGMA`, and `Borrower.B`, value `a` may equal precisely 0. This could cause division by 0 in `LiquidityAmounts` or elsewhere.

Things seem safe, but a bit tenuous. Would be good to make safer.

haydenshively:
Fixes #66


core/.storage-layout.md
```diff
@@ -26,6 +26,8 @@ forge inspect --pretty src/Factory.sol:Factory storage-layout
 |---------------|---------------------------------------------------------------|------|--------|-------|-------------------------|
 | getMarket     | mapping(contract IUniswapV3Pool => struct Factory.Market)     | 0    | 0      | 32    | src/Factory.sol:Factory |
 | getParameters | mapping(contract IUniswapV3Pool => struct Factory.Parameters) | 1    | 0      | 32    | src/Factory.sol:Factory |
-| isBorrower    | mapping(address => bool)                                      | 2    | 0      | 32    | src/Factory.sol:Factory |
-| couriers      | mapping(uint32 => struct Factory.Courier)                     | 3    | 0      | 32    | src/Factory.sol:Factory |
+| isLender      | mapping(address => bool)                                      | 2    | 0      | 32    | src/Factory.sol:Factory |
+| isBorrower    | mapping(address => bool)                                      | 3    | 0      | 32    | src/Factory.sol:Factory |
+| couriers      | mapping(uint32 => struct Factory.Courier)                     | 4    | 0      | 32    | src/Factory.sol:Factory |
+| isCourier     | mapping(address => bool)                                      | 5    | 0      | 32    | src/Factory.sol:Factory |
```
haydenshively, line 32: Unrelated. I just forgot to run the layout script a while back, and this file has been out of date ever since.

core/src/libraries/BalanceSheet.sol
```diff
@@ -92,7 +91,7 @@ library BalanceSheet {
             isSus = metric > _manipulationThreshold(_effectiveCollateralFactor(sigma));
 
             a = uint160((sqrtMeanPriceX96 * SoladyMath.sqrt(1e18 - sigma)) / 1e9);
-            b = SafeCastLib.safeCastTo160((sqrtMeanPriceX96 * SoladyMath.sqrt(1e18 + sigma)) / 1e9);
+            b = uint160(SoladyMath.min((sqrtMeanPriceX96 * SoladyMath.sqrt(1e18 + sigma)) / 1e9, type(uint160).max));
```
haydenshively, line 94: Instead of reverting when `b` exceeds uint160, we now clamp to it. This ensures massive prices won't brick the contract.

core/src/libraries/LiquidityAmounts.sol
```diff
@@ -3,66 +3,13 @@ pragma solidity 0.8.17;
...
-            // Store the function selector of `FullMulDivFailed()`.
-            mstore(0x00, 0xae47f702)
-            // Revert with (offset, size).
-            revert(0x1c, 0x04)
-        }
-
-        // Divide [prod1 prod0] by 2^96.
-        result := or(shr(96, prod0), shl(160, prod1))
-    }
-}
```
haydenshively, line 29: Moved to new file.

core/src/libraries/LiquidityAmounts.sol
```diff
@@ -3,66 +3,13 @@ pragma solidity 0.8.17;
...
-    function getAmount1ForLiquidity(
-        uint160 sqrtRatioAX96,
-        uint160 sqrtRatioBX96,
-        uint128 liquidity
-    ) internal pure returns (uint256 amount1) {
-        assert(sqrtRatioAX96 <= sqrtRatioBX96);
-
-        amount1 = mulDiv96(liquidity, sqrtRatioBX96 - sqrtRatioAX96);
-    }
-
```
haydenshively, line 65: These were only used in this file, so I marked them private. To conform with styleguide, moved to bottom of file. And since `getAmountsForLiquidity` already checks `sqrtRatioAX96 < sqrtRatioBX96`, I was able to remove the extra checks.

core/src/libraries/Volatility.sol
```diff
@@ -167,9 +167,9 @@ library Volatility {
         assert(sqrtRatioAX96 <= sqrtRatioX96 && sqrtRatioX96 <= sqrtRatioBX96);
 
         unchecked {
-            uint256 numerator = mulDiv96(sqrtRatioX96, sqrtRatioBX96 - sqrtRatioX96);
+            uint256 numerator = Math.mulDiv(uint256(liquidity) << 128, sqrtRatioX96, sqrtRatioBX96);
 
-            value0 = Math.mulDiv(liquidity, numerator, sqrtRatioBX96);
+            value0 = mulDiv224(numerator, sqrtRatioBX96 - sqrtRatioX96);
             value1 = mulDiv96(liquidity, sqrtRatioX96 - sqrtRatioAX96);
```
haydenshively, line 173: Should match the 2nd branch of the version of this function that's in `LiquidityAmounts`. Copied here because in this case, we don't care about the other branches.

IanWoodard (approved):

PR #141, opened 2023-08-31
Move liquidation ante payout to after _saveSlot

haydenshively:
Better checks/effects/interactions


aneesh1 (approved):

IanWoodard (approved):
You may want to split this big of a change up into multiple PRs in the future 🔢

PR #142, opened 2023-08-31
Rename sigma constants and organize other ones

haydenshively:
While integrating contracts on the frontend, @IanWoodard pointed out that the naming of IV / sigma / sigmaN was inconsistent and confusing. Hopefully this fixes that.

Also went ahead and organized the constants file a bit with headers, and replaced the in-line 8's in `Factory` with an actual constant.


core/.gas-snapshot
```diff
@@ -9,7 +9,7 @@ BorrowerGasTest:test_uniswapDepositStandard() (gas: 167558)
 BorrowerGasTest:test_uniswapWithdraw() (gas: 149326)
 BorrowerGasTest:test_withdraw() (gas: 107760)
 FactoryGasTest:test_createBorrower() (gas: 156437)
-FactoryGasTest:test_createMarket() (gas: 3882077)
```
IanWoodard, line 12: 😮‍💨

core/src/libraries/constants/Constants.sol
```diff
@@ -31,27 +35,45 @@ uint216 constant DEFAULT_ANTE = 0.1 ether;
...
+//////////////////////////////////////////////////////////////*/
+
+/// @dev The minimum implied volatility. Clamped to this **before** multiplying by `nSigma`.
+/// Expressed as a wad percentage at `IV_SCALE`, e.g. {0.01e18, 24 hours} → 1% daily → 19% annual
+uint256 constant IV_MIN = 0.01e18;
+
+/// @dev The maximum implied volatility. Clamped to this **before** multiplying by `nSigma`.
+/// Expressed as a wad percentage at `IV_SCALE`, e.g. {0.18e18, 24 hours} → 18% daily → 344% annual
+/// To avoid underflow in `BalanceSheet.computeProbePrices`, ensure that `IV_MAX * nSigma <= 1e18`
+uint256 constant IV_MAX = 0.18e18;
```
IanWoodard, line 75: Is there a reason we call it `IV_MIN` and `IV_MAX` rather than `MIN_IV` and `MAX_IV`

IanWoodard (approved):

core/src/libraries/constants/Constants.sol
haydenshively: Yes. It's not this way yet, but goal is to have the suffix be all that changes to indicate min, max, or default. Makes them look nice when importing alphabetically. Some of the other constants don't follow this pattern yet but I'll deal with them soon.

PR #143, opened 2023-09-01
Reduce sqrt usage in Volatility library

haydenshively:
Improved gas efficiency of Volatility library by doing 1 sqrt instead of 3. Price we pay for this is that it _may_ revert (depending on price) if fee growth exceeds 2^96 in the 6 hour interval, instead of 2^128.

This is the difference between 10 billion DAI/ETH/etc and like 100 quintillion DAI/ETH/etc. So our safety factor vs. current volumes goes from 50 trillion to 5000. Imo safety factor of 5000 is still plenty, especially given that the failure mode doesn't prevent people from withdrawing/repaying.


core/test/libraries/Volatility.t.sol
```diff
@@ -177,14 +177,14 @@ contract VolatilityTest is Test {
...
-        tickTVL = Volatility.computeTickTVLX64(60, -3000, TickMath.getSqrtRatioAtTick(-3000), 999999999);
-        assertEq(tickTVL, 47558380999913911951032320);
+        tickTVL = Volatility.computeTickTVL(1, 19000, TickMath.getSqrtRatioAtTick(19000), 100000000000);
+        assertEq(tickTVL, 12926964);
+        tickTVL = Volatility.computeTickTVL(10, 19000, TickMath.getSqrtRatioAtTick(19000), 9763248618769789);
+        assertEq(tickTVL, 12618077941403);
+        tickTVL = Volatility.computeTickTVL(60, -19000, TickMath.getSqrtRatioAtTick(-19000), 100000000000);
+        assertEq(tickTVL, 115925394);
+        tickTVL = Volatility.computeTickTVL(60, -3000, TickMath.getSqrtRatioAtTick(-3000), 999999999);
+        assertEq(tickTVL, 2578145);
```
haydenshively, line 187: The new numbers are the old numbers, divided by 2^64.

We're not actually losing precision here, we're just no longer bit shifting (which added 64 bits of fake "precision" before)

core/test/libraries/Volatility.t.sol
```diff
@@ -177,14 +177,14 @@ contract VolatilityTest is Test {
...
-        tickTVL = Volatility.computeTickTVLX64(60, -3000, TickMath.getSqrtRatioAtTick(-3000), 999999999);
-        assertEq(tickTVL, 47558380999913911951032320);
+        tickTVL = Volatility.computeTickTvl(1, 19000, TickMath.getSqrtRatioAtTick(19000), 100000000000);
+        assertEq(tickTVL, 12926964);
+        tickTVL = Volatility.computeTickTvl(10, 19000, TickMath.getSqrtRatioAtTick(19000), 9763248618769789);
+        assertEq(tickTVL, 12618077941403);
+        tickTVL = Volatility.computeTickTvl(60, -19000, TickMath.getSqrtRatioAtTick(-19000), 100000000000);
+        assertEq(tickTVL, 115925394);
+        tickTVL = Volatility.computeTickTvl(60, -3000, TickMath.getSqrtRatioAtTick(-3000), 999999999);
+        assertEq(tickTVL, 2578145);
```
haydenshively: New values are just the old ones, divided by 2^64. Not actually losing precision here, just not adding fake precision via bit shifts

core/src/libraries/Volatility.sol
```diff
@@ -130,47 +128,47 @@ library Volatility {
...
+                TickMath.getSqrtRatioAtTick(tick + tickSpacing),
+                liquidity
+            );
+        }
     }
 
     /**
      * @notice Computes the value of the liquidity in terms of token1
-     * @dev Each return value can fit in a uint192 if necessary
+     * @dev The return value can fit in uint193 if necessary
```
IanWoodard, line 151: Isn't a uint193 not even a thing in solidity?

IanWoodard (approved):

core/src/libraries/Volatility.sol
haydenshively: Yeah, but I thought it was more informative to say uint193 than uint200 (since 200 would be more than enough)

PR #144, opened 2023-09-05
Move payment logic to its own fn as prep for max repay

haydenshively:


aneesh1 (approved):

core/src/Lender.sol
```diff
@@ -518,4 +514,14 @@ contract Lender is Ledger {
         lastBalance = cache.lastBalance.safeCastTo112();
         lastAccrualTime = uint32(block.timestamp); // Disables reentrancy guard if there was one
     }
+
+    function _ensurePayment(uint256 expectedBalance) private {
+        unchecked {
```
IanWoodard, line 519: Why unchecked? Gas savings?

IanWoodard (approved):

Issue #58, opened 2023-01-26
Portions of `Lender.borrow` could be unchecked


haydenshively:
I'm fairly certain the following lines of `Lender.borrow` could be `unchecked`:
```
cache.borrowBase += units;
borrows[msg.sender] = b + units;
```

haydenshively:
Actually only the 2nd line could be unchecked. Not worth the messy code.

PR #145, opened 2023-09-06
Change repay require statement so max repay is always possible

haydenshively:
In order to recover their `ante` from a `Borrower`, user's must repay **all** debts. This has always been possible, but because of rounding errors there was no way to guarantee it could be done in a single call. Even if you had specified `borrowBalance` as the repay amount (which is, as the name implies, supposed to be your entire borrow balance) you could sometimes be left with units of debt.

This change fixes that.

I may (separately) add a helper function called `repayMax` that just calls `repay` under the hood, with `borrowBalance(...)` as the amount. Haven't decided yet whether this should go in `Lender` or the periphery's `Router`


core/src/Lender.sol
```diff
@@ -208,12 +208,14 @@ contract Lender is Ledger {
...
+        unchecked {
+            // Convert `amount` to `units`
+            units = (amount * BORROWS_SCALER) / cache.borrowIndex;
 
-        // Track borrows
+            // Track borrows
+            borrows[msg.sender] = b + units;
+        }
         cache.borrowBase += units;
-        borrows[msg.sender] = b + units;
```
haydenshively, line 216: https://github.com/aloelabs/aloe-ii/issues/58

Changing the function anyway at this point, figured I might as well go back and do this.

core/src/Lender.sol
```diff
@@ -208,12 +208,14 @@ contract Lender is Ledger {
         // Accrue interest and update reserves
         (Cache memory cache, ) = _load();
 
-        // Convert `amount` to `units`
-        units = amount.mulDivUp(BORROWS_SCALER, cache.borrowIndex);
+        unchecked {
+            // Convert `amount` to `units`
+            units = (amount * BORROWS_SCALER) / cache.borrowIndex;
```
haydenshively, line 213: Two reasons why it's okay to _not_ round up:
- As shown in the new `test_math_borrowUnitsUniqueness` test, `BORROWS_SCALER` is large enough that any change in `amount` will cause a corresponding change in `units`, and `units` is never 0 unless `amount` also is. In other words, we're not losing precision
- There's no free money glitch, because the `units` in the `repay` function are computed in the exact same way. So if you borrow some `amount`, the only way to repay all `units` of debt is to repay that same `amount`.

Ok, but why did I bother changing it at all?

Well one of the key attributes of this PR is that `lender.borrowBalance` is more precise. If you call `repay(borrowBalance(someAddr), someAddr)`, it is now guaranteed that the debt will be zeroed out -- previously it was just kinda close-ish. I had to choose floor/ceil very carefully to get this to work.

core/src/Lender.sol
```diff
@@ -235,7 +237,12 @@ contract Lender is Ledger {
         unchecked {
             // Convert `amount` to `units`
             units = (amount * BORROWS_SCALER) / cache.borrowIndex;
-            require(units < b, "Aloe: repay too much");
+            if (!(units < b)) {
+                units = b - 1;
+
+                uint256 maxRepay = (units * cache.borrowIndex).unsafeDivUp(BORROWS_SCALER);
+                require(b > 1 && amount <= maxRepay, "Aloe: repay too much");
+            }
```
haydenshively, line 245: `b - 1` is the maximum value we can subtract from the `borrows` mapping because there needs to at least 1 unit left as a flag that indicates the user is, in fact, a borrower.

However we can't simply clamp to that in all cases. Imagine a user sends $1000000 to the contract and calls this function with that amount, thinking they have that much debt. But in reality they have only $100 of debt. If we clamped to `b - 1` in all cases, `repay` would succeed and the remaining $999900 would be wasted (and likely stolen).

To protect users, we ensure the `amount <= maxRepay`. This is the same sort of protection we had before this PR, but now it's compatible with the whole "If you pay off `borrowBalance` you're guaranteed to be left with 0 debt" thing.

core/test/invariants/LenderHarness.sol
```diff
@@ -511,9 +511,11 @@ contract LenderHarness {
...
 
         address beneficiary = borrowers[i % count];
-        // uint256 amount = LENDER.borrowBalance(beneficiary);
-        // if (amount > type(uint112).max) amount = type(uint112).max;
-        return repay(uint112(LENDER.borrowBalance(beneficiary)), beneficiary);
+        uint256 units = repay(uint112(LENDER.borrowBalance(beneficiary)), beneficiary);
+
+        require(LENDER.borrows(beneficiary) == 1, "repay: didn't repay max units");
+        require(LENDER.borrowBalance(beneficiary) == 0, "repay: didn't repay max amount");
+        return units;
```
haydenshively, line 518: Too lax before. Now we assert that borrows is 1 (so just the is-borrower flag) and borrowBalance is 0.

IanWoodard (approved):

aneesh1 (approved):

PR #151, opened 2023-09-09
Add governance methods in Factory

haydenshively:
Open questions:
- I've set it up so nSigma can be governed between 4 and 5. Is that even worth it? We could add precision to it (so governance can control more decimal points between 4 and 5) or remove it altogether.
- Since raising nSigma from 4 to 5 could cause people to become liquidatable instantly, should we consider building a timelock into the Factory? i.e., the decision to raise nSigma must be made like a week before it actually goes through. Note that governance itself is likely to have a timelock, so putting one in Factory could be considered redundant. But if this code is to be self-sufficient, there's still an argument for it.

Other potential things that could be governable (currently constants):
- manipulation threshold divisor
- iv change per second


aneesh1 (approved):

core/src/Lender.sol
```diff
@@ -50,23 +50,32 @@ contract Lender is Ledger {
...
         initialDomainSeparator = _computeDomainSeparator();
         initialChainId = block.chainid;
+    }
 
+    /// @notice Sets the `rateModel` and `reserveFactor`. Only the `FACTORY` can call this.
+    function setRateModelAndReserveFactor(IRateModel rateModel_, uint8 reserveFactor_) external {
+        require(msg.sender == address(FACTORY));
         rateModel = rateModel_;
-        require(MIN_RESERVE_FACTOR <= reserveFactor_ && reserveFactor_ <= MAX_RESERVE_FACTOR);
         reserveFactor = reserveFactor_;
```
IanWoodard, line 66: Should we emit an event after setting the rate model and reserve factor?

core/src/Lender.sol
```diff
@@ -50,23 +50,32 @@ contract Lender is Ledger {
...
 
         initialDomainSeparator = _computeDomainSeparator();
         initialChainId = block.chainid;
+    }
 
+    /// @notice Sets the `rateModel` and `reserveFactor`. Only the `FACTORY` can call this.
+    function setRateModelAndReserveFactor(IRateModel rateModel_, uint8 reserveFactor_) external {
+        require(msg.sender == address(FACTORY));
         rateModel = rateModel_;
-        require(MIN_RESERVE_FACTOR <= reserveFactor_ && reserveFactor_ <= MAX_RESERVE_FACTOR);
```
IanWoodard, line 62: I know we check this in the factory, which is the only place this can be called, but it still feels weird not having a check in here for added safety (I know it costs a little extra gas).

IanWoodard (approved):

core/src/Lender.sol
haydenshively: Yes. Good catch; I'll make sure to add this when I add the timelock (1 or 2 PR's down the line)

core/src/Lender.sol
haydenshively: I agree it feels a little weird, but I want to keep all the `CONSTRAINT_XXX` constants in the Factory so that it's clear to people what exactly is allowed and what isn't. If we apply constraints in multiple places, that gets fuzzier.

PR #156, opened 2023-09-11
Fix pausing functionality

haydenshively:
The `if (isSus) FACTORY.pause` line was kinda useless before, because if `isSus` was true, it would revert. Probably should've been caught in the audit, but at least I found it now.

I removed that line from `modify` (and reorganized that stuff a bit). Now to do a `pause`, anyone can call the function in the `Factory`. It'll double check whether the prices are legit or not, and if they aren't, it'll pause the market.


core/test/Utils.sol
```diff
@@ -14,11 +14,6 @@ import {Lender} from "src/Lender.sol";
 import {RateModel} from "src/RateModel.sol";
 import {VolatilityOracle} from "src/VolatilityOracle.sol";
 
-function deploySingleBorrower(IUniswapV3Pool pool, Lender lender0, Lender lender1) returns (Borrower) {
-    address oracleMock = address(new VolatilityOracleMock());
-    return new Borrower(VolatilityOracle(oracleMock), pool, lender0, lender1);
-}
-
```
haydenshively, line 21: Unused code

core/test/libraries/BalanceSheet.t.sol
```diff
@@ -12,121 +12,155 @@ import {
...
+        uint256 liabilities1
+    ) external pure returns (bool) {
+        return BalanceSheet.isHealthy(prices, mem, liabilities0, liabilities1);
+    }
+}
+
 contract BalanceSheetTest is Test {
     function setUp() public {}
 
+    function test_alwaysHealthyWhenLiabilitiesAre0(
```
haydenshively, line 33: Since I moved the isHealthy check inside of the `liabilities0 > 0 || liabilities1 > 0` conditional, I wanted to double check that no debt does, in fact, imply you're healthy. This test proves that.

aneesh1 (approved):
LGTM. Also, looks like we have some [_major_ gas savings](https://github.com/aloelabs/aloe-ii/pull/156/files#diff-44ae0a2517e37bc4f980e6ed336771c0e8f8b5ede5d38f6ab16fd7f942a18e5cR6) on `repay`!

core/src/Borrower.sol
```diff
@@ -439,18 +437,21 @@ contract Borrower is IUniswapV3MintCallback {
...
         (, uint256 nSigma, ) = FACTORY.getParameters(UNISWAP_POOL);
-        (prices, ) = _getPrices(nSigma, oracleSeed);
+        (prices, seemsLegit) = _getPrices(oracleSeed, nSigma);
     }
 
-    function _getPrices(uint256 nSigma, uint40 oracleSeed) private view returns (Prices memory prices, bool isSus) {
+    function _getPrices(
+        uint40 oracleSeed,
+        uint256 nSigma
+    ) private view returns (Prices memory prices, bool seemsLegit) {
```
IanWoodard, line 448: isSus => seemsLegit, great naming

core/src/Borrower.sol
```diff
@@ -295,21 +295,19 @@ contract Borrower is IUniswapV3MintCallback {
...
 
-        require(BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: unhealthy");
-        unchecked {
-            if (liabilities0 + liabilities1 > 0)
-                require(
-                    address(this).balance > ante && !isSus && block.timestamp > pausedUntilTime,
-                    "Aloe: missing ante / sus price"
-                );
+            require(
+                seemsLegit && (block.timestamp > pausedUntilTime) && (address(this).balance >= ante),
```
IanWoodard, line 304: Why go from using gt to using gte?

core/src/Borrower.sol
```diff
@@ -295,21 +295,19 @@ contract Borrower is IUniswapV3MintCallback {
...
+                "Aloe: missing ante / sus price"
+            );
+
+            Assets memory assets = _getAssets(positions_, prices, false);
+            require(BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: unhealthy");
```
IanWoodard, line 309: Moved to the end

core/src/Borrower.sol
```diff
@@ -295,21 +295,19 @@ contract Borrower is IUniswapV3MintCallback {
         int24[] memory positions_ = positions.write(callee.callback(data, msg.sender));
         _saveSlot0(uint160(msg.sender), _formatted(State.Ready));
 
-        (uint256 ante, uint256 nSigma, uint256 pausedUntilTime) = FACTORY.getParameters(UNISWAP_POOL);
-
-        (Prices memory prices, bool isSus) = _getPrices(nSigma, oracleSeed);
-        Assets memory assets = _getAssets(positions_, prices, false);
         (uint256 liabilities0, uint256 liabilities1) = _getLiabilities();
+        if (liabilities0 > 0 || liabilities1 > 0) {
```
IanWoodard, line 299: Moved to the top

core/src/libraries/BalanceSheet.sol
```diff
@@ -85,10 +85,10 @@ library BalanceSheet {
         uint256 iv,
         uint256 nSigma,
         uint56 metric
-    ) internal pure returns (uint160 a, uint160 b, bool isSus) {
+    ) internal pure returns (uint160 a, uint160 b, bool seemsLegit) {
         unchecked {
             iv = SoladyMath.clamp((nSigma * iv) / 10, PROBE_PERCENT_MIN, PROBE_PERCENT_MAX);
-            isSus = metric > _manipulationThreshold(_effectiveCollateralFactor(iv));
+            seemsLegit = metric < _manipulationThreshold(_effectiveCollateralFactor(iv));
```
IanWoodard, line 91: Should this be `<=`?

core/src/Factory.sol
```diff
@@ -103,8 +103,10 @@ contract Factory {
         DEFAULT_RATE_MODEL = defaultRateModel;
     }
 
-    function pause(IUniswapV3Pool pool) external {
-        require(isBorrower[msg.sender]);
+    function pause(IUniswapV3Pool pool, uint40 oracleSeed) external {
```
IanWoodard, line 106: I am still confused about when/how pause is run. Does it have to be called manually? If so, what is stopping someone from not calling it?

IanWoodard (approved):

core/src/Factory.sol
aneesh1: Anyone can call `pause`. So nothing is stopping anyone from calling it. However, if they call it and the conditions are healthy, then based on [this](https://github.com/aloelabs/aloe-ii/pull/156/files/37333e5bab43db4512330181b965b678d893fc89#diff-fcdc0cc0ffdbed060a1297c96a2c6786bc6ad73d015a0ae3ae18d3841e8362bbR107-R109) I don't think any pausing actually happens.

core/src/Borrower.sol
haydenshively: Thank you, I strive for perfection :)

core/src/Borrower.sol
haydenshively: So that we don't have to add the 1 wei at the end. Nbd in frontend code, but if someone ever wants to fund a `Borrower` manually, it's much easier to type "0.01" than "0.010000000...00001"

core/src/libraries/BalanceSheet.sol
haydenshively: Nah it doesn't really matter. The manipulation stuff isn't an exact science. This way's more conservative and saves a few gas.

PR #161, opened 2023-09-18
Fix LTV symmetry and add Borrower tests

haydenshively:
While writing new tests for the `Borrower` I noticed an inconsistency. When computing probe prices, we essentially did `currentPrice * (1 + somePercent)` and `currentPrice * (1 - somePercent)`, i.e. linearly adding/subtracting percentages from the mean.

Though this _is_ a common way of forecasting future price ranges using IV [[1]](https://www.optionsanimal.com/using-implied-volatility-determine-expected-range-stock/), it leads to an asymmetry in LTV. If you have {collateral0, borrow1}, you end up with the nominal LTV that we had in the constants file. But if you have {collateral1, borrow0}, you end up with a slightly different LTV. This isn't exactly an attack vector, but as @IanWoodard would say, it feels yucky.

To fix this, we now compute probe prices like `currentPrice * thing` and `currentPrice / thing`, with `thing = e ** (nSigma * IV)`. You can see the difference between the old and new way on [this Desmos](https://www.desmos.com/calculator/xoczvlacij). [This thread](https://quant.stackexchange.com/questions/54812/probability-of-a-stock-price-using-implied-volatility) and [this ChatGPT answer](https://chat.openai.com/share/83d172fd-84df-4344-8c18-28d8ce36c6d6) are why I chose that particular definition of `thing`.

After fixing that, I was still seeing a small asymmetry in the LTVs. Revisiting some old math, I found that it was caused by our liquidation incentive calculation. Even when evaluating health at probe price A or B, we evaluate liquidation incentive at the TWAP. This prevents liquidators from using IV manipulation to inflate their reward, but means that our health check sometimes had `liabilities0` multiplied by both price A/B _and_ the TWAP in different places.

Solved by evaluating the liquidation incentive at probe prices, _at least for health-check purposes_. So no more mixing and matching price A/B and the TWAP. However, to retain the manipulation protection properties, the _actual payout_ is still always computed at TWAP.

TLDR: There were two subtle math issues causing effective LTV to differ from my expectation. Fixed both of them and addressed other repercussions (like manipulation threshold tweaks you see in `BalanceSheet.t.sol`


core/src/Borrower.sol
```diff
@@ -189,28 +189,26 @@ contract Borrower is IUniswapV3MintCallback {
...
             // Calculate liquidation incentive
-            (incentive1, priceX128) = BalanceSheet.computeLiquidationIncentive(
+            incentive1 = BalanceSheet.computeLiquidationIncentive(
                 assets.fixed0 + assets.fluid0C, // total assets0 at `prices.c` (the TWAP)
                 assets.fixed1 + assets.fluid1C, // total assets1 at `prices.c` (the TWAP)
                 liabilities0,
                 liabilities1,
-                prices.c
+                priceX128
             );
```
haydenshively, line 209: It just felt silly to be calculating `priceX128` inside `computeLiquidationIncentive`. "Each function should do one thing and one thing only" bla bla bla

core/src/Borrower.sol
```diff
@@ -189,28 +189,26 @@ contract Borrower is IUniswapV3MintCallback {
...
             // Ensure only unhealthy accounts can be liquidated
-            require(!BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1, incentive1), "Aloe: healthy");
+            require(!BalanceSheet.isHealthy(prices, assets, liabilities0, liabilities1), "Aloe: healthy");
```
haydenshively, line 211: `isHealthy` does its own evaluation of the liquidation incentive, at `prices.a` and `prices.b` (instead of passing in the one above, which was found at `prices.c`)

core/src/libraries/BalanceSheet.sol
```diff
@@ -26,34 +33,21 @@ struct Prices {
...
-        // don't want to re-evaluate it at the probe prices (as would happen if we added it to `liabilities0`).
         unchecked {
-            liabilities0 += liabilities0 / MAX_LEVERAGE;
-            liabilities1 += liabilities1 / MAX_LEVERAGE + incentive1;
+            liabilities0 +=
+                (liabilities0 / MAX_LEVERAGE) +
+                (liabilities0.zeroFloorSub(mem.fixed0 + mem.fluid0C) / LIQUIDATION_INCENTIVE);
+            liabilities1 +=
+                (liabilities1 / MAX_LEVERAGE) +
+                (liabilities1.zeroFloorSub(mem.fixed1 + mem.fluid1C) / LIQUIDATION_INCENTIVE);
```
haydenshively, line 50: Note the liquidation incentive is still a bit of a hybrid calc, because we're measuring shortfall amounts at the TWAP:

`max(0, liabilitiesX - (mem.fixedX + mem.fluidXC))`

instead of at prices A and B using something like `fluidXA` and `fluidXB`. That's okay though, because for this fix, it doesn't matter _where the amounts come from_ so long as the price we use to combine them is uniform (see remainder of this fn).

core/src/libraries/BalanceSheet.sol
```diff
@@ -76,17 +70,21 @@ library BalanceSheet {
 
     function computeProbePrices(
         uint56 metric,
-        uint160 sqrtMeanPriceX96,
+        uint256 sqrtMeanPriceX96,
```
haydenshively, line 73: Still expect it to be uint160 coming in, this is just to avoid annoying typecasting inside the fn.

core/src/libraries/BalanceSheet.sol
```diff
@@ -76,17 +70,21 @@ library BalanceSheet {
...
+            uint256 sqrtScaler = uint256(exp1e12(int256(nSigma * iv) / 20)).clamp(
+                PROBE_SQRT_SCALER_MIN,
+                PROBE_SQRT_SCALER_MAX
+            );
 
-            a = uint160((sqrtMeanPriceX96 * SoladyMath.sqrt(1e12 - iv)) / 1e6);
-            b = uint160(SoladyMath.min((sqrtMeanPriceX96 * SoladyMath.sqrt(1e12 + iv)) / 1e6, type(uint160).max));
+            seemsLegit = metric < _manipulationThreshold(_ltv(sqrtScaler), manipulationThresholdDivisor);
+
+            a = uint160((sqrtMeanPriceX96 * 1e12).rawDiv(sqrtScaler));
```
IanWoodard: Clamp this to be no smaller than sqrtRatioMIn

core/src/libraries/BalanceSheet.sol
```diff
@@ -95,11 +93,9 @@ library BalanceSheet {
         uint256 assets1,
         uint256 liabilities0,
         uint256 liabilities1,
-        uint160 sqrtMeanPriceX96
-    ) internal pure returns (uint256 incentive1, uint256 meanPriceX128) {
+        uint256 meanPriceX128
+    ) internal pure returns (uint256 incentive1) {
         unchecked {
-            meanPriceX128 = square(sqrtMeanPriceX96);
```
IanWoodard, line 101: ❤️

core/src/libraries/constants/Constants.sol
```diff
@@ -94,24 +94,24 @@ uint256 constant LIQUIDATION_INCENTIVE = 20;
 /// involves the swap callbacks (`swap1For0` and `swap0For1`). There is no grace period for in-kind liquidations.
 uint256 constant LIQUIDATION_GRACE_PERIOD = 2 minutes;
 
+/// @dev TODO:
```
IanWoodard: TODO?

core/src/libraries/constants/Constants.sol
```diff
@@ -94,24 +94,24 @@ uint256 constant LIQUIDATION_INCENTIVE = 20;
...
+uint256 constant PROBE_SQRT_SCALER_MIN = 1.026248453011e12;
+
+/// @dev TODO:
```
IanWoodard: TODO?

IanWoodard (approved):

PR #163, opened 2023-09-20
Write tests for non-governance Factory functionality

haydenshively:


core/src/Factory.sol
```diff
@@ -87,11 +92,6 @@ contract Factory {
                            INCENTIVE STORAGE
     //////////////////////////////////////////////////////////////*/
 
-    struct Courier {
-        address wallet;
-        uint16 cut;
-    }
-
```
IanWoodard, line 94: This big of a change should have been split up into multiple PRs 🌵

core/test/Factory.t.sol
```diff
@@ -0,0 +1,159 @@
...
+        emit EnrollCourier(id, address(this), cut);
+        factory.enrollCourier(id, cut);
+
+        assertTrue(factory.isCourier(address(this)));
+        (address courier, uint16 cutRec) = factory.couriers(id);
+        assertEq(courier, address(this));
+        assertEq(cutRec, cut);
+    }
+
+    function test_claimRewards() external {
```
IanWoodard, line 114: I feel like this isn't a very exhaustive test

IanWoodard (approved):

PR #164, opened 2023-09-21
Clean up / finish Factory

haydenshively:
I decided not to implement a timelock here, as the only parameter it matters for is nSigma, and whatever `GOVERNOR` is implemented down the road will likely have a timelock built in.

I did add event emission when parameters are changed, and added comments / organized stuff.


core/src/libraries/LiquidityAmounts.sol
```diff
@@ -110,7 +110,7 @@ library LiquidityAmounts {
     /// @param sqrtRatioBX96 A sqrt price representing the second tick boundary
     /// @param liquidity The liquidity being valued
     /// @return amount1 The amount of token1. Will fit in a uint192 if you need it to
-    function getAmount1ForLiquidity(
+    function _getAmount1ForLiquidity(
```
haydenshively, line 113: linter

core/src/Borrower.sol
```diff
@@ -111,8 +111,7 @@ contract Borrower is IUniswapV3MintCallback {
         TOKEN0 = lender0.asset();
         TOKEN1 = lender1.asset();
 
-        require(pool.token0() == address(TOKEN0));
-        require(pool.token1() == address(TOKEN1));
+        assert(pool.token0() == address(TOKEN0) && pool.token1() == address(TOKEN1));
```
IanWoodard, line 114: What is the difference between assert and require?

core/src/Factory.sol
```diff
@@ -40,60 +39,95 @@ contract Factory {
...
+        // The time at which borrowing can resume
         uint32 pausedUntilTime;
     }
 
-    struct Courier {
-        address wallet;
-        uint16 cut;
-    }
-
+    // The set of all governable `Market` properties
```
IanWoodard, line 66: Why does this keep getting moved??

IanWoodard (approved):

core/src/Factory.sol
haydenshively: In the previous PR it was sitting amongst the state variables, so I moved it up. Here it's just being shifted down because I added a new struct

core/src/Borrower.sol
haydenshively: require is usually used for checking external inputs, assert is for internal stuff. Since this data is coming from the Factory, I'm considering it internal.

I debated taking out the check entirely too.

PR #169, opened 2023-09-25
Clean up periphery

haydenshively:
- Linting (function ordering especially)
- Moved certain files to https://github.com/aloelabs/aloe-ii-examples
- Updated BorrowerLens to match the rounding-up we do in `BalanceSheet`
- Withdraw ante when exiting a Boosted NFT position


periphery/src/LenderLens.sol
```diff
@@ -18,13 +17,28 @@ contract LenderLens {
...
+     * @dev In most cases, `lender.maxRedeem(owner)` returns a static number of shares, and a standard `lender.redeem`
+     * call can successfully redeem everything. However, if the user has a courier or if utilization is too high, the
+     * number of shares may change block by block. In that case, to redeem the maximum value, you can pass
+     * `type(uint256).max` to `lender.redeem`.
+     */
+    function isMaxRedeemDynamic(Lender lender, address owner) external view returns (bool) {
+        // NOTE: If the first statement is true, the second statement will also be true (unless this is the block in which
+        // they deposited for the first time). We include the first statement only to reduce computation.
+        return lender.courierOf(owner) > 0 || lender.balanceOf(owner) != lender.maxRedeem(owner);
     }
```
haydenshively, line 43: This property is pretty `Lender`-centric, so I thought it made more sense here than in the `Router`

periphery/src/LenderLens.sol
```diff
@@ -18,13 +17,28 @@ contract LenderLens {
...
+            uint8 reserveFactor
         )
     {
         asset = lender.asset();
         (, inventory, totalBorrows, totalSupply) = lender.stats();
 
-        if (inventory != 0) utilization = Math.mulDiv(1e18, totalBorrows, inventory);
+        if (inventory > 0) utilization = (1e18 * totalBorrows) / inventory;
         interestRate = lender.rateModel().getYieldPerSecond(utilization, address(lender));
+        reserveFactor = lender.reserveFactor();
```
haydenshively, line 29: Return reserveFactor as part of `readBasics`. I checked frontend code; since this is at the end of the array it shouldn't break any of our current stuff. Just makes it possible to simplify calls in the future.

periphery/src/boost/NFTDescriptor.sol
```diff
@@ -95,19 +95,19 @@ library NFTDescriptor {
                 " Uniswap V3 pool.\\n",
                 "\\nPool: ",
                 poolString,
-                "\\nPosition: ",
+                "\\n\\nBorrower: ",
```
haydenshively, line 98: OpenSea seems not to render newlines unless there are two in a row 🤷‍♂️

PR #171, opened 2023-09-27
Allow owner of Borrower to re-initialize to change owner

haydenshively:
The `BoostNFT` owns/controls a bunch of `Borrowers` on behalf of NFT owners. By default, the `boostManager` is used as the `IManager` in all `modify` callbacks. If a user dislikes what the `boostManager` is doing or wants additional functionality, they can override it with a custom `IManager`. However, the `BoostNFT` will always be the owner.

It'd be nice to have a more complete escape hatch from things like the `BoostNFT` and the upcoming `MultiBorrower` I have in mind for Fuse 2. Ownership transfer is that escape hatch.

The only real downside is that checking for `CreateBorrower` events on the `Factory` is no longer sufficient to resolve all `Borrower`s for a given user (since someone may have transferred ownership to them). UX / dev ex tradeoff.


core/src/Borrower.sol
```diff
@@ -116,8 +116,11 @@ contract Borrower is IUniswapV3MintCallback {
 
     receive() external payable {}
 
+    /// @dev This allows re-initialization to a different `owner`
     function initialize(address owner) external {
-        require(slot0.owner == address(0));
+        uint256 ownerAndState = _loadSlot0() & 0xff0000000000000000000000ffffffffffffffffffffffffffffffffffffffff;
```
haydenshively, line 121: Masks out the `unleashLiquidationTime`. Lowest 160 bits are the `owner`, and highest 8 are the `state` flag.

We want `state == State.Ready`, which is 0. So we should be left with just the address, and can assert that it's either 0 or msg.sender.

aneesh1: A bit confused. According to this [article](https://medium.com/@mweiss.eth/solidity-and-evm-bit-shifting-and-masking-in-assembly-yul-942f4b4ebb6a), `f = 1111`. 

```
struct Slot0 {
    address owner;
    uint88 unleashLiquidationTime;
    State state;
}
```
Shouldn't the mask be `0xffffffffffffffffffffffffffffffffffffffff0000000000000000000000ff`?

aneesh1 (commented):
One question about the mask.

core/src/Borrower.sol
haydenshively: See the following snippet from the article; in our case `state` is like variable `F`, so it should be first in our mask.

image

PR #174, opened 2023-10-01
Adding foundry gas diff gh action

IanWoodard:


.github/workflows/foundry-gas-diff.yml
```diff
@@ -0,0 +1,61 @@
...
+    # Optionally configure to run only for changes in specific files. For example:
+    # paths:
+    # - src/**
+    # - test/**
+    # - foundry.toml
+    # - remappings.txt
+    # - .github/workflows/foundry-gas-diff.yml
+
+jobs:
+  compare_gas_reports:
```
haydenshively: nit: I think we use dashes in our other actions, not underscores

.github/workflows/foundry-gas-diff.yml
```diff
@@ -0,0 +1,61 @@
...
+    defaults:
+      run:
+        working-directory: 'core'
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install Foundry
+        uses: onbjerg/foundry-toolchain@v1
```
haydenshively: Why this fork instead of `foundry-rs/foundry-toolchain@v1`

.github/workflows/foundry-gas-diff.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          version: nightly
+
+      # Add any step generating a gas report to a temporary file named gasreport.ansi. For example:
+      - name: Run tests
+        run: ./snapshot.sh > ../gasreport.ansi # <- this file name should be unique in your repository!
```
haydenshively: Instead of trying to move to gasreport.ansi (which I'm not sure this will accomplish, since the script doesn't have any stdout)

Simply set the `report` in the Rubilmax action (https://github.com/Rubilmax/foundry-gas-diff#report-string)

.github/workflows/foundry-gas-diff.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          # make fuzzing semi-deterministic to avoid noisy gas cost estimation
+          # due to non-deterministic fuzzing (but still use pseudo-random fuzzing seeds)
+          FOUNDRY_FUZZ_SEED: 0x${{ github.event.pull_request.base.sha || github.sha }}
+
+      - name: Compare gas reports
+        uses: Rubilmax/foundry-gas-diff@v3.16
```
haydenshively, line 46: On the other hand, it looks like this expects a different format (the one from `forge test --gas-report`) instead of the one from `forge snapshot`. So you may need to revert to what you had before, but with an added `--match-contract ".*Gas"` flag)

i.e. in the "Run tests" step, use `forge test --gas-report --match-contract ".*Gas" > ../gasreport.ansi`

haydenshively (commented):
This would be really nice to have but I don't think it's working yet.

.github/workflows/foundry-gas-diff.yml
IanWoodard: That doesn't work since the path name can't have a forward slash in it 🌵

.github/workflows/foundry-gas-diff.yml
IanWoodard: I don't know. I just used the one they use.

IanWoodard: It looks like they are the same thing

haydenshively: Probably use the new one that it redirects to then

haydenshively (approved):
Seems legit. Let's try it out and we can finetune from there.

PR #203, opened 2023-10-29
Add courier enrollment fee to mitigate griefing vector

haydenshively:
Require prospective couriers to pay 0.01 ETH in order to enroll. Filling all 2^32 - 1 `courierId`s would now cost nearly 43 million ETH, regardless of gas costs. Front-running individual enrollment transactions is also an order of magnitude more expensive, but if it's still a problem, users could enroll via a private mempool.

Addresses
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/1
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/26
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/56
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/140

### DIFF
--- a/core/script/Deploy.s.sol
+++ b/core/script/Deploy.s.sol
@@ -9,7 +9,7 @@ import {RateModel} from "src/RateModel.sol";
 import {VolatilityOracle} from "src/VolatilityOracle.sol";
 
 address constant GOVERNOR = 0x2bf8eA41aF0695D482C4aa23d4f8aD7E9023b890;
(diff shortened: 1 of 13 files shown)


aneesh1 (approved):

haydenshively:
Closing because it's infeasible on mainnet and current L2's. If we want to deploy to a super cheap L2 in the future we can reexamine this.

PR #214, opened 2023-11-08
Reduce manipulation attack surface of Volatility library

haydenshively:
The IV formula requires us to compute the value of liquidity at the "current" tick. Mathematically, the most correct way of doing that is to look at `pool.slot0().sqrtPriceX96` and `pool.liquidity()`, which is what we did previously. However, those values are manipulatable, and it turns out that using the geometric mean price and the harmonic mean liquidity give approximately the same results. Moreover, since we were already using these mean values to get Fee Volume in the numerator, they end up cancelling out and simplifying the math.

The IV estimate is now based purely on `feeGrowthGlobals` and the TWAP. Since `feeGrowthGlobals` are themselves manipulatable (albeit at a cost), further work could include reducing `IV_CHANGE_PER_UPDATE`, using a median fee growth value over many samples, or both.

Addresses
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/63
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/109
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/125

image

Other IV issues that this does **not** yet address
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/38
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/45
- https://github.com/sherlock-audit/2023-10-aloe-judging/issues/91

### DIFF
--- a/core/src/VolatilityOracle.sol
+++ b/core/src/VolatilityOracle.sol
@@ -43,25 +43,19 @@ contract VolatilityOracle {
     }
 
     function update(IUniswapV3Pool pool, uint40 seed) external returns (uint56, uint160, uint256) {
-        // TODO: require that sample for the current block hasn't been written to Uniswap observations,
-        // implying that nobody has added a bunch of liquidity to manipulate the sample
         unchecked {
             // Read `lastWrite` info from storage
             LastWrite memory lastWrite = lastWrites[pool];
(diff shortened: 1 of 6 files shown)


haydenshively:
Note: expect merge conflicts in `Oracle.sol` with #217

PR #231, opened 2023-12-15
Deploy contracts using solc 0.8.23 for proper verification

haydenshively:
0.8.17 has a bug which makes compiling non-deterministic depending on the solc backend, causing my local bytecode (compiled on macOS) to differ from Etherscan's (presumably linux-based compiler) and fail verification.

Also revamped the deploy scripts to avoid use of the experimental `--multi` flag.


core/foundry.toml
```diff
@@ -7,10 +7,11 @@ libs = ['lib']
 out = 'build'
 cache_path = 'cache'
 
-solc_version = '0.8.17'
+solc_version = '0.8.23'
+evm_version = 'paris'
 via_ir = true
 optimizer = true
-optimizer_runs = 4294967295
+optimizer_runs = 100000000
```
haydenshively, line 14: This is the maximum supported by Etherscan's JSON parser.

